PR #14, opened 2018-05-06
added pytest for show_all_columns and match_rates arguments

ryanshiroma:
I forgot to copy the code for the show_all_columns functionality in my previous commit! I've added that code back in and added another pytest to test both show_all_columns and match_rates.


theianrobertson:
LGTM!  I'll cut a version with this and previous changes, just need to remember how to deploy to pypi 😁

PR #18, opened 2018-05-30
strip leading and trailing blanks in strings

### LINKED ISSUE #15: Would be useful to have a parameter to strip spaces for comparison
As probably expected, the following code will return a mismatch since 'B'<>'B ':
```
import pandas as pd
import datacompy

df1 = pd.DataFrame([
        {'id': 1234, 'column_value': 'A'},
        {'id': 2345, 'column_value': 'B'}])

df2 = pd.DataFrame([
        {'id': 1234, 'column_value': 'A'},
        {'id': 2345, 'column_value': 'B '}])

compare = datacompy.Compare(
    df1,
    df2,
    join_columns='id',
    abs_tol=0,
    rel_tol=0,
    )
compare.matches(ignore_extra_columns=False)
# False

# This method prints out a human-readable report summarizing and sampling differences
print(compare.report())
```
What I propose is an optional parameter to ignore differences where the only difference is leading or trailing spaces. In this example it is obvious that there is a trailing space. However, when we are dealing with extracts from different databases/source files, without real control over the ETL of these, sometimes we can't prevent these discrepancies. We may wish to ignore these types of mismatches to identify 'worse' mismatches more effectively.

Another candidate could be ignoring case sensitivity differences.
Of course these could both be easily handled with preprocessing the dataframes, but still could be some convenient enhancements!

fdosani:
@theianrobertson 

- Some logic to strip leading and trailing blanks in strings. Fixes #15
- Should be noted that this won't work on string columns which are used as `join_columns` 
  (custom index).
   - will still lead to mismatches on that specific case
   - depending on how you want to tackle this it could be something worth discussing.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -59,6 +59,8 @@ class Compare(object):
         more easily track the dataframes.
     df2_name : str, optional
         A string name for the second dataframe
+    ignore_spaces : bool, optional
+        Flag to strip whitespace (including newlines) from string columns
 
     Attributes
     ----------
@@ -70,7 +72,7 @@ class Compare(object):
 
     def __init__(
         self, df1, df2, join_columns=None, on_index=False, abs_tol=0,
-        rel_tol=0, df1_name='df1', df2_name='df2'):
+        rel_tol=0, df1_name='df1', df2_name='df2', ignore_spaces=False):
 
         if on_index and join_columns is not None:
             raise Exception('Only provide on_index or join_columns')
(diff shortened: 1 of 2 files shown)


theianrobertson:
:shipit:

I confirmed that the tests pass on Pandas 0.20.3, so failure on Travis should be OK, pending resolution of #19

theianrobertson:
(Thanks for the contribution btw!)

PR #30, opened 2018-09-12
Ignore the case of a string when comparing. 

### LINKED ISSUE #29: Ignore case when comparing strings
Might be a nice to have a optional ignore case feature so that `STRING` == `string`.
Should be relatively simple to add. Happy to take it on.

@theianrobertson thoughts?

fdosani:
Closes #29 

@theianrobertson feel free to ignore this PR if it isn't appropriate.
Was relatively easy and seemed like something which might make sense for some use cases.


theianrobertson:
:shipit:

PR #36, opened 2019-01-23
Closes #35 by fixing dedupe bug

### LINKED ISSUE #35: Issue with nulls in join columns when de-duping
There's an issue with the logic that tries to handle duplicate data.  This is a minimal example:

```python
import numpy as np
import pandas as pd
import datacompy

df1 = pd.DataFrame({
    'fld_1': [1,2,2,3,3,4,5,5],
    'fld_2': ['A',np.nan,np.nan,np.nan,np.nan,np.nan,np.nan,np.nan]})

df2 = pd.DataFrame({
    'fld_1': [1,2,3,4,5],
    'fld_2': ['A',np.nan,np.nan,np.nan,np.nan]})

comp = datacompy.Compare(df1, df2, join_columns=['fld_1', 'fld_2'])
assert comp.subset()
```

`df2` is a subset of `df1`, but it's not matching.  The issue is [here](https://github.com/capitalone/datacompy/blob/master/datacompy/core.py#L220), the `groupby.cumcount` just lumps any rows with nulls in them into one big category.  Need to make that more resilient to nulls.

theianrobertson:
Blah not sure if there's a better way to solve this, but I can't figure out a cleaner way to handle nulls when trying to assign a unique value within a group where the group by columns can have nulls in them.  @fdosani mind taking a peek?  Happy to chat through the issue if you'd like.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -772,26 +772,16 @@ def generate_id_within_group(dataframe, join_columns):
     Pandas.Series
         The ID column that's unique in each group.
     """
-    default_date = datetime(1985, 6, 20)
-    default_other = -999
-
-    # Temporarily fill join column values with default values
-    temp_joiners = dataframe.sort_values(by=list(dataframe.columns))[join_columns].copy()
-    for column in join_columns:
-        if temp_joiners[column].isnull().any():
-            if np.issubdtype(temp_joiners[column].dtype, np.datetime64):
-                if (temp_joiners[column] == default_date).any():
-                    raise ValueError(
-                        "Could not deduplicate null rows with value {}".format(default_date)
-                    )
-                else:
-                    temp_joiners[column].fillna(default_date, inplace=True)
-            else:
-                if (temp_joiners[column] == default_other).any():
-                    raise ValueError(
-                        "Could not deduplicate null rows with value {}".format(default_other)
-                    )
-                else:
-                    temp_joiners[column].fillna(default_other, inplace=True)
-
-    return temp_joiners.groupby(join_columns).cumcount()
+    default_value = "DATACOMPY_NULL"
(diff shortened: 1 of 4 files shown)


theianrobertson:
I wasn't really thrilled with the first version of this, but I think I'm happier with this version after sleeping on it 😁

fdosani:
Taking a look 👍

datacompy/core.py
```diff
@@ -761,3 +754,34 @@ def calculate_max_diff(col_1, col_2):
...
+        The dataframe to operate on
+    join_columns : list
+        List of strings which are the join columns
+
+    Returns
+    -------
+    Pandas.Series
+        The ID column that's unique in each group.
+    """
+    default_value = "DATACOMPY_NULL"
```
fdosani: Is the expectation this is passed in with the DF? I'm probably missing the context around this, but where does `DATACOMPY_NULL` get set or come from?

theianrobertson: This is to handle cases where there are both nulls and duplicates in the columns you're joining on.

I was using `df.groupby(join_columns).cumcount()` to generate an ID that was unique within each unique group, but because of how groupby drops nulls, anything with a null in a join_column would just get lumped into one big group.

To handle that case we need to `fillna` before `groupby`.  In this PR, if there are nulls in the join columns, I convert all the join columns to text, then just fill nulls with `DATACOMPY_NULL`, which (hopefully) shouldn't exist in your dataframe.  If it does exist it'll fail, and the user will have to fillna themselves before comparing.

Guess I should probably add something in the docs too :)

fdosani: Makes sense!
Yeah a blurb in the docs might be helpful for the context around this.

Issue #13, opened 2018-05-01
Align the APIs between Compare and SparkCompare


theianrobertson:
Not sure if it makes sense to go all the way to subclassing or ABCs, but the API calls between `Compare` and SparkCompare` are quite different.  I think they could be aligned somewhat before adding any new functionality.

theianrobertson:
Have been poking around with some preliminary stuff here in the `combine` branch.  To break this down a little bit:

- [x] Change names on SparkCompare to df1/df2 instead of base/compare
- [x] Move non-compare functions out of core.py into utils.py
- [x] Move core.py to use the default stdout print, one report function that calls a bunch of others (text into templates too)
- [ ] Add in tests on the report rendering for Compare - right now it just checks that it renders at all
- [ ] Convert sparkcompare report to follow the same pattern as Compare - mostly renaming, but may have to pull the merge step out to match with Compare
- [ ] Make sparkcompare report out the dataset names
- [ ] Add in column translations for Compare?
- [ ] Add in sampling for SparkCompare report
- [ ] redefine SparkCompare as `class SparkCompare(Compare)`
- [ ] (maybe) align testing across Pandas and Spark?  Like have tests that should generate the same output in both?

dan-coates:
@fdosani and @theianrobertson, this was a long-standing bugaboo of mine, and I think I'm going to put some time into maybe finishing this off. Since there are some judgment calls in this sort of alignment, I wanted to check to see if there are any changes that have been made so far or are planned that are fundamental here, vs. what's still open for discussion.

I'll weigh in with further thoughts once I look at what's happened so far and what's yet to be done, but wanted to orient myself to what you've already thought of here.

dan-coates:
I just went through a lengthy conflict resolution to make `combine` up to date with `master` so I got a little taste of the changes... my main question off the bat is whether or not we should use ABCs here. I tend to think we should, and here's why:

`Compare` is a pretty large beast. It defines many internal and external methods and properties - it looks like there may be 50 of them. Simply understanding what may be in one class but not the other will likely require a spreadsheet and manually ticking off attributes in both classes.

[...]

PR #40, opened 2019-05-08
Report combine

theianrobertson:
This is going to the `combine` branch.  I'm pulling the trigger on having `SparkCompare` inherit from `Compare`, and moving to a unified report generation process where some function calls are abstracted and implemented separately in both, then called from the same report functions that are in `Compare`.

The `Compare` branch isn't "done" yet - there's still a few things in #13 that I will still add before cutting a new major release.

I'm also enabling the Spark unit tests on travis instead of just locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,3 +5,7 @@ repos:
     - id: black
       language_version: python3.7
       types: [python]
+-   repo: https://github.com/pre-commit/mirrors-isort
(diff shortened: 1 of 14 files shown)


theianrobertson:
Hey @fdosani you mind taking a lil' peek at this when you have a minute?  I've been poking away at this for a bit without any other eyes on it, probably better to give it a little bit of light :grin:

PR #49, opened 2019-09-25
Travis fix

fdosani:
@jborchma if you don't mind taking a look too. 
This should fix the TravisCI issue, and clean up a few things in the tests and in particular `pytest` and how it deals with `raises`


jborchma:
lgtm

Issue #48, opened 2019-09-25
Dropping support for Python 2?


fdosani:
Wondering if we should have some conversation about dropping Python 2 support as it is coming soon to EOL. Right now we are support both, but it might help simplify things if we only go with 3.6+ in the next major release?

@theianrobertson @jborchma @gforsyth @slott56 Would appreciate any feedback on this.

jborchma:
Let's drop it!

fdosani:
I agree 👍

gforsyth:
Strong agree. Drop it

hugovk:
Please see PR #51.

PR #51, opened 2019-09-27
Drop support for Python 2

hugovk:
Fixes https://github.com/capitalone/datacompy/issues/48.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 jdk: openjdk8
 dist: trusty
-sudo: false
 cache:
   pip: true
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ dist: trusty
 cache:
   pip: true
 python:
-  - "2.7"
   - "3.6"
 before_install:
   - mkdir -p /opt
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: trusty
 cache:
   pip: true
 python:
+  - "3.5"
   - "3.6"
 before_install:
   - mkdir -p /opt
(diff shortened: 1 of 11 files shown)


fdosani:
@hugovk Thank you for this PR. Will take a look in a bit. In the mean time (if you haven't already) could you please complete the [CLA](https://docs.google.com/forms/d/19LpBBjykHPox18vrZvBbZUcK6gQTj7qv1O5hCduAZFU/viewform)

hugovk:
Done!

requirements.txt
```diff
@@ -1,4 +1,3 @@
 enum34>=1.1.6;python_version<"3.4"
```
fdosani: thinking we should just remove `enum` here too. I don't think we want to be supporting older versions of 3 either. Feel like at the minimum 3.5+

.travis.yml
```diff
@@ -1,11 +1,9 @@
 language: python
 jdk: openjdk8
 dist: trusty
-sudo: false
 cache:
   pip: true
 python:
-  - "2.7"
```
fdosani: Might make sense to swap this out for:

```yaml
  - "3.5"
  - "3.6"
  - "3.7"
```

hugovk: Good idea.

However, Python 3.7 requires `dist: xenial` on Travis, and a quick test shows more work is needed to test against Xenial, so I'll only add 3.6 as part of this PR and 3.7 can be dealt with in another issue.

https://travis-ci.org/hugovk/datacompy/builds/591518867

fdosani: Sounds good! Thank you Hugo.

hugovk:
Updated!

PR #54, opened 2020-01-06
Add datacompy[spark] pip install option

dan-coates:
This change allows a user to run `pip install datacompy[spark]` and install datacompy along with its prerequisite for Spark usage (pyspark). Right now, `pip install datacompy` comes with requirements needed for pandas comparisons, but not Spark. That is probably the correct choice to not inconvenience pandas users with a heavy Spark dependency, but forces Spark users to manually account for the dependency. This change intends to be a reasonable compromise of giving users the choice of a one-step installation for datacompy with Spark if desired.

I also included a few housekeeping changes that were easy to make as I ran through the code:
* Update copyright year to present
* Sort imports per best practices
* Run pre-commit black and Travis on Python 3.7*
* Avoid deprecated `unionAll` method in SparkCompare tests
* Eliminate Python 2.7 compatibility code in setup.py
* Pull package requirements from requirements.txt in setup.py to avoid the possibility of the two becoming misaligned
* Minor black-compatible formatting changes in a couple of places

\*I'm not sure if this will cause any CI failures, but the only way to know is to try :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,16 @@
 from setuptools import setup, find_packages
 import os
-from io import open as iopen
 
 CURR_DIR = os.path.abspath(os.path.dirname(__file__))
-INSTALL_REQUIRES = ["pandas>=0.19.0,!=0.23.*", "numpy>=1.11.3"]
-with iopen(os.path.join(CURR_DIR, "README.rst"), encoding="utf-8") as file_open:
+
+with open(os.path.join(CURR_DIR, "README.rst"), encoding="utf-8") as file_open:
     LONG_DESCRIPTION = file_open.read()
+
+with open("requirements.txt", "r") as requirements_file:
+    raw_requirements = requirements_file.read().strip().split("\n")
+
+INSTALL_REQUIRES = [line for line in raw_requirements if not (line.startswith("#") or line == "")]
+
 
 exec(open("datacompy/_version.py").read())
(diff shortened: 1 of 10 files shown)


fdosani:
@dan-coates for some reason I never got a notice for this. Sorry about the huge delay with this Dan. I'll try and review the PR shortly.

setup.py
```diff
@@ -18,6 +23,7 @@
     license="Apache-2.0",
     packages=find_packages(),
     install_requires=INSTALL_REQUIRES,
+    extras_require={"spark": ["pyspark>=2.2.0"]},
```
fdosani, line 26: This would essentially install the latest `pyspark` but that won't match with what the user potentially might have installed. I guess users might need to change this after the fact. Thoughts?

dan-coates: Well, this says we need >= v2.2. If a user includes this in their requirements.txt with pyspark==2.4.0, I think our requirement won't prevent the user's installation from fetching the version of pyspark that they want, but this probably depends on their package management (i.e. I'm not 100% sure how pip, pipenv, conda, poetry would handle this situation).

fdosani: I agree. Some stuff will be up to end users to manage, and Spark is a bit weird as the pip version corresponds to the actual Spark version. LGTM. Approving.

PR #62, opened 2020-05-14
Fixes #59

### LINKED ISSUE #59: Doesnt take into account join columns when doing strip() of white spaces
Join columns can have whitespaces as well,so ideally below section should be enhanced to check for ignore_spaces and strip() if it is true for string columns
This is causing complete mismatch of 2 dataframes here in case 1 of join column has space:

if on_index and join_columns is not None:
            raise Exception("Only provide on_index or join_columns")
        elif on_index:
            self.on_index = True
            self.join_columns = []
        elif isinstance(join_columns, str):
            self.join_columns = [join_columns.lower()]
            self.on_index = False
        else:
            self.join_columns = [col.lower() for col in join_columns]
            self.on_index = False

fdosani:
@jborchma This fixes issues #59 
Just want your thoughts on this. It is a pretty simple fix described below:

Under `_dataframe_merge` we check if `ignore_spaces` is `true`, and apply the `strip` on the `join_colmuns` if it is a `O` dtype object. This is done for `df1` and `df2`

When some calls: `datacompy.Compare(df1, df2, "a", ignore_spaces=False)`
It will ignore spaces essentially on all string columns including the join ones.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -231,12 +231,11 @@ class Compare:
             params = {"on": self.join_columns}
 
         if ignore_spaces:
-            self.df1[self.join_columns] = self.df1[self.join_columns].apply(
-                lambda x: x.str.strip() if x.dtype.kind == "O" else x
-            )
-            self.df2[self.join_columns] = self.df2[self.join_columns].apply(
-                lambda x: x.str.strip() if x.dtype.kind == "O" else x
-            )
+            for column in self.join_columns:
+                if self.df1[column].dtype.kind == "O":
+                    self.df1[column] = self.df1[column].str.strip()
+                if self.df2[column].dtype.kind == "O":
+                    self.df2[column] = self.df2[column].str.strip()
 
         outer_join = self.df1.merge(
             self.df2, how="outer", suffixes=("_df1", "_df2"), indicator=True, **params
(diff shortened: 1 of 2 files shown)


datacompy/core.py
```diff
@@ -229,6 +230,14 @@ def _dataframe_merge(self, ignore_spaces):
         else:
             params = {"on": self.join_columns}
 
+        if ignore_spaces:
+            self.df1[self.join_columns] = self.df1[self.join_columns].apply(
+                lambda x: x.str.strip() if x.dtype.kind == "O" else x
```
jborchma, line 235: Hmm, so `apply` is not the fastest...I guess you did this due to the element dtype check. How come you're doing that one? Maybe we could do it vectorized? All the `str` methods work as vectorized methods.

fdosani:
Yeah just for the check really. I don’t want to strip something which isn’t
a string as it will throw an error.

datacompy/core.py
fdosani: Yeah just for the check really. I don’t want to strip something which isn’t
a string as it will throw an error.

fdosani: I guess I can loop through the list check the dtype and then just vertorize right? Assuming that is what you were thinking?

fdosani: @jborchma Made a change to move away from `apply`. Does this feel better. Agree apply is known to be slow, and this could lead to some savings.

Issue #55, opened 2020-01-06
Use type hinting & support Python 3.5+


dan-coates:
Since we target Python 3.5+ in our CI testing, we can make this explicit and commit to only supporting that version and higher. This will enable us to use type hinting in our code, which would have a few benefits:

* IDE support for types while developing datacompy
* IDE support for types for users using datacompy
* Ability to do static type checking in our CI pipeline
* Automatic type inference during documentation generation

This should probably best be done in concert with something like #13 as long as we're messing with the code, but since this will involve committing to supporting certain versions of Python, it should be a conscious choice. Python 3.4 [reached EOL](https://devguide.python.org/devcycle/#end-of-life-branches) in March 2019 so I don't see a problem with explicitly targeting 3.5+.

fdosani:
@dan-coates I might look into implementing this for the next release. I think we are away from Python 2 support so the minimum moving forward will be 3.5+. did you want to take a crack? Or if you had other thoughts around this?

jborchma:
I would actually propose to drop 3.5 support as well. A lot of major packages have dropped support for that version and the official EOL is not far off.

fdosani:
Personally I think it makes sense to wait and do it once it is Officially
dead dead. Also I don’t think there is a huge cost at the moment to support
it.

jborchma:
True, EOL for 3.5 is in September.

fdosani:
Closing this as 3.5 is EOL and no longer part of the support plan.

PR #65, opened 2020-05-20
Type hints

jborchma:
In this PR I am adding type hints, partially addressing #55.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -109,9 +109,9 @@ class Compare:
         self.df2_name = df2_name
         self.abs_tol = abs_tol
         self.rel_tol = rel_tol
-        self.df1_unq_rows: pd.DataFrame
-        self.df2_unq_rows: pd.DataFrame
-        self.intersect_rows: pd.DataFrame
+        self.df1_unq_rows: Optional[pd.DataFrame] = None
+        self.df2_unq_rows: Optional[pd.DataFrame] = None
+        self.intersect_rows: Optional[pd.DataFrame] = None
         self.column_stats: List[Dict[str, Any]] = []
         self._compare(ignore_spaces, ignore_case)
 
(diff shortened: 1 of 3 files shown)


jborchma:
Oh, and this is of course breaking Python 3.5 support...

fdosani:
> Oh, and this is of course breaking Python 3.5 support...

Seems to be failing on the Spark side and for all the versions.

jborchma:
Ah, I think it's fixed now.

datacompy/core.py
```diff
@@ -108,16 +109,18 @@ def __init__(
         self.df2_name = df2_name
         self.abs_tol = abs_tol
         self.rel_tol = rel_tol
-        self.df1_unq_rows = self.df2_unq_rows = self.intersect_rows = None
-        self.column_stats = []
+        self.df1_unq_rows: pd.DataFrame
```
fdosani, line 112: I think this and the next 2 need to be fixed also?

```
Optional[Type["pd.DataFrame"]] = None
```

jborchma: So those are now only the python 3.5 tests that are failing due to how I used type hints here (this was added in 3.6)

fdosani: OK, but I think we still want to support Python 3.5 for now. so we need to figure out a way to get the tests to pass given the type hinting changes. Anyone running on Python 3.5 will hit issues I'm guessing at runtime.

jborchma: Yeah, I'll fix it. It will be some effort and will probably involve me just ignoring a bunch of rows for mypy.

jborchma: The main issue is that mypy doesn't recognize the logic that set the variable (from `None` to a dataframe) and then complains when a dataframe method is used (because `None` of course doesn't have those).

jborchma: So all the methods that use, for example, `self.intersect_rows` fail mypy because it could be `None`. The solution to this that I used works only starting with 3.6. Other solutions would be to check in every method if the variable is `None` and in that case error out (which would technically prevent those bugs that mypy finds from happening) or ignore the types for all those rows (29 in total).

jborchma: Any preference?

jborchma: I opted for raising `TypeError` in case any of the attributes are `None`.

jborchma:
Ah damn, I think we'll need to close this until we can drop  3.5 support...

Issue #68, opened 2020-06-22
`Compare._validate_dataframe` lowercases column names prior to checking for uniqueness


ferenyx:
Prior to the check for uniqueness, `_validate_dataframe` does `dataframe.columns = [col.lower() for col in dataframe.columns]`. If columns have unique mixed-case labels, these could collide when lower-cased as in this example:

```In [29]: df = pd.DataFrame(np.random.randn(2, 2), index=np.arange(0.0, 1.0, 0.5), columns=["OSs", "Oss"])

In [30]: df
Out[30]:
          OSs       Oss
0.0 -0.438774 -0.853128
0.5  0.919049 -0.795371

In [31]: x = datacompy.Compare(df, df, on_index=True)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-31-69a8e3399fdd> in <module>
----> 1 x = datacompy.Compare(df, df, on_index=True)

lib/python3.6/site-packages/datacompy/core.py in __init__(self, df1, df2, join_columns, on_index, abs_tol, rel_tol, df1_name, df2_name, ignore_spaces, ignore_case)
    103
    104         self._any_dupes = False
--> 105         self.df1 = df1
    106         self.df2 = df2
    107         self.df1_name = df1_name

lib/python3.6/site-packages/datacompy/core.py in df1(self, df1)
    121         """Check that it is a dataframe and has the join columns"""
    122         self._df1 = df1
--> 123         self._validate_dataframe("df1")
    124
    125     @property

lib/python3.6/site-packages/datacompy/core.py in _validate_dataframe(self, index)
    151
    152         if len(set(dataframe.columns)) < len(dataframe.columns):
--> 153             raise ValueError("{} must have unique column names".format(index))
    154
    155         if self.on_index:

ValueError: df1 must have unique column names```

ferenyx:
This is related to #67, and the fix could potentially be the same - please let me know if you'd prefer I report this as part of that issue!

jborchma:
I think adding a toggle to the lower casing might solve this. However, I would probably argue to keep the default to always do the lower casing.

ferenyx:
that sounds reasonable to me!

Issue #67, opened 2020-06-22
`Compare._validate_dataframe` assumes columns have string labels


ferenyx:
It appears that `_validate_dataframe` assumes that columns have string labels (it unconditionally calls `.lower()` on them), which raises `AttributeError` if columns instead have other types of labels (e.g. numerical):

```
In [18]: df = pd.DataFrame(np.random.randn(2), index=np.arange(0.0, 1.0, 0.5),
    ...:                             columns=[0.01])

In [19]: df
Out[19]:
         0.01
0.0 -0.283165
0.5 -0.388745

In [20]: compare = datacompy.Compare(df, df, on_index=True)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-20-d3440cd3839e> in <module>
----> 1 compare = datacompy.Compare(df, df, on_index=True)

lib/python3.6/site-packages/datacompy/core.py in __init__(self, df1, df2, join_columns, on_index, abs_tol, rel_tol, df1_name, df2_name, ignore_spaces, ignore_case)
    103
    104         self._any_dupes = False
--> 105         self.df1 = df1
    106         self.df2 = df2
    107         self.df1_name = df1_name

lib/python3.6/site-packages/datacompy/core.py in df1(self, df1)
    121         """Check that it is a dataframe and has the join columns"""
    122         self._df1 = df1
--> 123         self._validate_dataframe("df1")
    124
    125     @property

lib/python3.6/site-packages/datacompy/core.py in _validate_dataframe(self, index)
    145             raise TypeError("{} must be a pandas DataFrame".format(index))
    146
--> 147         dataframe.columns = [col.lower() for col in dataframe.columns]
    148         # Check if join_columns are present in the dataframe
    149         if not set(self.join_columns).issubset(set(dataframe.columns)):

lib/python3.6/site-packages/datacompy/core.py in <listcomp>(.0)
    145             raise TypeError("{} must be a pandas DataFrame".format(index))
    146
--> 147         dataframe.columns = [col.lower() for col in dataframe.columns]
    148         # Check if join_columns are present in the dataframe
    149         if not set(self.join_columns).issubset(set(dataframe.columns)):

AttributeError: 'float' object has no attribute 'lower'
```

jborchma:
We'll take a look!

PR #69, opened 2020-06-22
Lower casing columns and casting as string

jborchma:
In this PR I am addressing both #68 and #67. I have implemented the following things:

I have added a new attribute called `lower` that will toggle casting the column names to lower. The default is `True`, so nothing should change unless specified.

Further, I have added that column names are cast to strings in the internal dataframes. 

I have also added testing for both features.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -105,6 +105,7 @@ class Compare:
             self.on_index = False
 
         self._any_dupes = False
+        self.lower = lower
         self.df1 = df1
         self.df2 = df2
         self.df1_name = df1_name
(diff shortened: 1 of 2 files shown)


ferenyx:
Thanks for the quick turnaround!

Would it be worth clarifying the difference between `ignore_case` and `lower` (or using a more descriptive name - e.g. `ignore_columns_case`) so it's clear that `ignore_case` only applies to contents of columns, and `lower` (or `ignore_columns_case`) only applies to the labels on columns?

jborchma:
Yeah, good point, I'll change the name to something more descriptive

PR #88, opened 2020-11-21
Bring Python current in Travis

gliptak:


jborchma (approved):
This makes sense to me. I think we could eventually take out 3.5 since we are past its EOL.

fdosani:
I'll need to take a closer look at this. The Spark tests are failing due to the Java version I think.

gliptak:
@jborchma might need to move up to Pyspark 3.x for Python 3.8+

PR #99, opened 2021-01-13
Fixing Spark 3 tests

### LINKED ISSUE #93: Tests are failing in Spark 3 with Python 3.7.x
Apparently the test are failing when using Spark 3 with Python 3.7.x. See #91.

fdosani:
Closes #93

Needed to change the order of some of the datset schemas in order for the tests to work correctly. Seems like order matters more in Spark 3.

### DIFF
--- a/tests/test_sparkcompare.py
+++ b/tests/test_sparkcompare.py
@@ -1377,25 +1377,25 @@ def test_rows_only_base_returns_a_dataframe_with_rows_only_in_base(spark, compar
     schema = StructType(
         [
             StructField("acct", LongType(), True),
-            StructField("dollar_amt", LongType(), True),
-            StructField("name", StringType(), True),
-            StructField("float_fld", DoubleType(), True),
             StructField("date_fld", DateType(), True),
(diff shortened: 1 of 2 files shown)


tests/test_sparkcompare.py
```diff
@@ -1665,84 +1761,84 @@ def test_rows_both_all_returns_all_rows_in_both_dataframes_for_differently_named
     expected_df = spark.createDataFrame(
         [
             Row(
+                accnt_purge=False,
                 acct=10000001234,
+                date_fld_base=datetime.date(2017, 1, 1),
+                date_fld_compare=datetime.date(2017, 1, 1),
+                date_fld_match=True,
```
jborchma, line 1768: So essentially the order here was breaking the tests?

fdosani: Yeah it has to do the with the schema and how the union function works. I guess before it would ignore the order and now it has to. I’ll dig into this more but at least we know spark 3 works 😃

jborchma (approved):
lgtm

PR #102, opened 2021-02-10
Suppress "Sample Rows" section in the output report if sample_count=0

mnizol:
Currently, if the sample_count parameter in the report() method is set to 0, the comparison report still prints an empty data frame object, which can be confusing to the end user.  This pull request is for a minor change to the report() method to suppress the "Sample Rows" section of the output report entirely if sample_count = 0.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -571,14 +571,15 @@ class Compare:
             ].to_string()
             report += "\n\n"
 
-            report += "Sample Rows with Unequal Values\n"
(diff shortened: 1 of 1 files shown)


fdosani (approved):
@mnizol Thanks very much for the PR, appreciate the contribution. This LGTM!
@elzzhu , @jborchma any objections? If not then I can merge it in once CICD is done.

jborchma (approved):
lgtm

PR #107, opened 2021-04-09
Make column order in reports determinitistic (alphabetical)

### LINKED ISSUE #81: Use deterministic output order for column-based reporting
One section of the output report is called "Sample Rows with Unequal Values". It prints each column with differences and then 10 rows of differences. However, I've noticed that this output changes with multiple runs - so multiple calls of `.report()` will give you output in a different order. As an example, if you have two dataframes with the columns A, B, and C, all of which have differences, then one run may give you results in the order of B, C, A, the second in C, A, B, etc.

This isn't a huge issue - it's just cosmetic. But I do think the output would be cleaner if it the ordering was deterministic. Moreover, it should probably match the ordering of the columns of the first dataframe, which is referred to as `df1` in the code.

This section of the report is generated by looping through `self.column_stats` and generated sample rows of differences. `self.column_stats` is generated from the method `_intersect_compare()`, which loops through `self.intersect_columns()`. And finally, the issue is in `self.interesect_columns()`, which uses `set` - an unordered data structure - and does an intersection.

https://github.com/capitalone/datacompy/blob/8d96429c9cf574f8d448a4e63301f0025ad467ee/datacompy/core.py#L212-L214

I'm thinking the solution to this problem is use an ordered set. [There are packages available on PyPi for this](https://pypi.org/project/ordered-set/).

KrishanBhasin:
Sort the column intersection before using them in inspections.

Fixes #81

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -297,7 +297,7 @@ class Compare:
         """
         LOG.debug("Comparing intersection")
         row_cnt = len(self.intersect_rows)
-        for column in self.intersect_columns():
+        for column in sorted(self.intersect_columns()):
             if column in self.join_columns:
                 match_cnt = row_cnt
                 col_match = ""
@@ -369,7 +369,7 @@ class Compare:
             Number of matching rows
         """
         match_columns = []
-        for column in self.intersect_columns():
+        for column in sorted(self.intersect_columns()):
             if column not in self.join_columns:
                 match_columns.append(column + "_match")
         return self.intersect_rows[match_columns].all(axis=1).sum()
--- a/tests/test_sparkcompare.py
+++ b/tests/test_sparkcompare.py
@@ -45,7 +45,7 @@ CACHE_INTERMEDIATES = True
 # (if we need to use these in other modules, move to conftest.py)
 @pytest.fixture(scope="module", name="spark")
 def spark_fixture():
-    spark = SparkSession.builder.master("local[2]").appName("pytest").getOrCreate()
+    spark = SparkSession.builder.master("local[2]").config("spark.driver.bindAddress", "127.0.0.1").appName("pytest").getOrCreate()
     yield spark
     spark.stop()
 


jborchma (commented):
Tests seem to be failing because of some Spark issue. In general I think this fix is great!

KrishanBhasin:
Thanks for the quick review @jborchma. I agree, the failures look Spark related (and spurious: `java.net.BindException: Cannot assign requested address: Service 'sparkDriver' failed after 16 retries (on a random free port)! Consider explicitly setting the appropriate binding address for the service 'sparkDriver' (for example spark.driver.bindAddress for SparkDriver) to the correct binding address.`)

What should we do to proceed? I don't have access to re trigger the CI to see if it resolves itself.

fdosani (changes requested):
Can you try changing this line: https://github.com/KrishanBhasin/datacompy/blob/550014c32637b98088e97bde78660fb6be653613/tests/test_sparkcompare.py#L48 to 

```
spark = SparkSession.builder.master("local[2]").config("spark.driver.bindAddress", "127.0.0.1").appName("pytest").getOrCreate()
```

Would like to see if that fixes it.

fdosani (approved):
Nice, that worked

datacompy/core.py
```diff
@@ -297,7 +297,7 @@ def _intersect_compare(self, ignore_spaces, ignore_case):
         """
         LOG.debug("Comparing intersection")
         row_cnt = len(self.intersect_rows)
-        for column in self.intersect_columns():
+        for column in sorted(self.intersect_columns()):
```
gandhis1, line 300: A sorted column order is deterministic, but I really think you ought to use an ordered set structure here, because doing so will maintain the order of the columns as per the source file. Also, when doing the intersection I imagine there is a way to do a "stable" intersection, akin to a stable sort, which will maintain the column order of the source insofar as the columns match.

Let's say I have a list of people, with the following columns:

|Name|DateOfBirth|Address|City|State|ZipCode|
----------------------------------------------------

And let's say for example's sake the other file being compared against has identical columns. I'd probably want to make the columns keep the same order, for readability, rather than reordering to this:

|Address|City|DateOfBirth|Name|State|ZipCode|
----------------------------------------------------

gandhis1: Here is the area that I think should use an ordered set and to the extent possible a "stable" intersection:

https://github.com/capitalone/datacompy/blob/8d96429c9cf574f8d448a4e63301f0025ad467ee/datacompy/core.py#L212-L214

KrishanBhasin: I was going for the simplest way to make the output deterministic without adding extra dependencies, as a quick way of obtaining an incremental improvement over the previous implementation. 

I think there are definitely improvements over the simple sorted list I have implemented; please open a PR with your suggested changes!

PR #108, opened 2021-04-09
Use OrderedSet to maintain DataFrame column order through set operations

### LINKED ISSUE #81: Use deterministic output order for column-based reporting

gandhis1:
@fdosani @KrishanBhasin 

Fixes #81. Iteration on #107. This aims to maintain the original order when printing out the set of common columns as well as unique columns, both which are based on set operations.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -461,6 +461,27 @@ def test_columns_no_overlap():
     assert compare.intersect_columns() == {"a", "b"}
 
 
+def test_columns_maintain_order_through_set_operations():
+    df1 = pd.DataFrame(
+        [
+            (("A"), (0), (1), (2), (3), (4), (-2)),
+            (("B"), (0), (2), (2), (3), (4), (-3))
+        ],
+        columns=["join", "f", "g", "b", "h", "a", "c"]
+    )
+    df2 = pd.DataFrame(
+        [
+            (("A"), (0), (1), (2), (-1), (4), (-3)),
+            (("B"), (1), (2), (3), (-1), (4), (-2))
+        ],
+        columns=["join", "e", "h", "b", "a", "g", "d"]
+    )
+    compare = datacompy.Compare(df1, df2, ["join"])
+    assert list(compare.df1_unq_columns()) == ["f", "c"]
+    assert list(compare.df2_unq_columns()) == ["e", "d"]
+    assert list(compare.intersect_columns()) == ["join", "g", "b", "h", "a"]
+
+
 def test_10k_rows():
     df1 = pd.DataFrame(np.random.randint(0, 100, size=(10000, 2)), columns=["b", "c"])
     df1.reset_index(inplace=True)
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -28,6 +28,8 @@ import os
 import numpy as np
 import pandas as pd
 
+from ordered_set import OrderedSet
+
 LOG = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


tests/test_core.py
```diff
@@ -461,6 +461,27 @@ def test_columns_no_overlap():
...
+        [
+            (("A"), (0), (1), (2), (-1), (4), (-3)),
+            (("B"), (1), (2), (3), (-1), (4), (-2))
+        ],
+        columns=["join", "e", "h", "b", "a", "g", "d"]
+    )
+    compare = datacompy.Compare(df1, df2, ["join"])
+    assert list(compare.df1_unq_columns()) == ["f", "c"]
+    assert list(compare.df2_unq_columns()) == ["e", "d"]
+    assert list(compare.intersect_columns()) == ["join", "g", "b", "h", "a"]
```
gandhis1, line 482: This example is intended to try to test edge cases (make sure the intersection is "stable" based on the left side), but doesn't necessarily reflect a realistic case. A realistic case is comparing two DataFrames where the columns are virtually identical with minimal differences, and it is in this context that maintaining the original DataFrame ordering actually becomes useful.

fdosani (commented):
Thanks for the PR @gandhis1. @KrishanBhasin previous PR #107 partially fixes this correct. This is just a variation on the implementation from what I am seeing? In principal I'm fine with the changes, its more adding another dependency which I'd like to discuss. The library was last touched in June 2020. @jborchma @elzzhu thoughts?

fdosani:
@gandhis1 Would you be open to implementing this without an external dependency? Maybe a `collections.OrderedDict` which just uses the keys. Talking about this with the team and I agree the order would be helpful but I'm a bit hesitant with introducing a new dependency like `ordered-set`

gandhis1:
So we can copy in the source code here, which removes the PyPi package dependency. The library is under MIT license which as I understand means we just need a copyright and license notice on the file.

The challenge with using `OrderedDict` is it apparently does not support ordered set operations:

```
In [1]: x = OrderedDict([("b", None), ("a", None), ("c", None)])                                        

In [2]: y = OrderedDict([("c", None), ("b", None), ("z", None)])                                        

In [3]: x.keys()                                                                                        
Out[3]: odict_keys(['b', 'a', 'c'])

In [4]: x.keys() - y.keys()                                                                            
Out[4]: {'a'}

In [5]: type(x.keys() - y.keys())                                                                      
Out[5]: set

In [6]: x.keys() | y.keys()                                                                            
Out[6]: {'a', 'b', 'c', 'z'}
```

gandhis1:
Revisiting this - any issue if I literally copy the original source code, which is a single file covered under the MIT license, into this repository? MIT license does not affect the license of any other code in this repository, as long as the original license is included with the file.

fdosani:
@SanthiSridharan sorry for the extended delay on this. Been really busy.
So I think we can merge this in. the library doesn't introduce any sub-dependencies and I think that is fine in itself. If this causes issues we can revisit it at a later date.

fdosani (approved):
LGTM

PR #138, opened 2022-04-08
Create whitesource.config

fdosani:
Adding in a new WS config file to use Python 3.9


fdosani:
WS is failing, but this is a known issue with the py37 usage vs higher.

PR #148, opened 2022-05-15
Spark compare refactor

### LINKED ISSUE #13: Align the APIs between Compare and SparkCompare

### LINKED ISSUE #147: Use "<=>" operator to join spark dataframes
If there are nulls in the join columns of dataframes, they do not pass the join condition and so are not counted as common rows, and instead counted as unique to each dataframe. If the "<=>" operator is used instead of "=" in these statements, null values will be joined:
            join_condition = " AND ".join(
                ["A." + name + "=B." + name for name in self._join_column_names]

This will make the spark compare behave more like the pandas compare, as null values join there. Note that "<=>" only works in later versions of Spark, so maybe it should be an option, or just make the change and require at least the necessary version (at least spark 1.6 according to the first answer in this: https://stackoverflow.com/questions/41728762/including-null-values-in-an-apache-spark-join ).

_Originally posted by @runrig in https://github.com/capitalone/datacompy/discussions/94#discussioncomment-2694968_

fdosani:
@dan-coates @theianrobertson I've taken a stab at this long outstanding issue #13. I just made the decision to follow the advice from the original discussion and make a `ABC` based on the pandas versions, and then implement the same set of items in Spark (or pretty close). It's a slightly involved PR, So appreciate some eyes on this. There are some differences which I should note:

- No indexing like pandas has it. Joining is done on `join_columns` like the old SparkCompare.
- Dups are removed (like the original SparkCompare and where it also differs from Pandas)
- In the new SparkCompare we are able to do a NullSafe joins now on the join columns. This was requested by @runrig in #147
-  also like the old SparkCompare there is no x-compare between non-numeric and numeric types. (Might be able to tackle this later, but the PR was getting to long). That also means we skip comparing strings and say dates, again something to add on to later.

Will start this off as a draft just to make sure folks are aligned. @runrig, if you have time to clone this branch and check out the new refactor I'd appreciate it. It should give you what you were looking for in `<=>` joins. You should be able to `pip install -e .[dev,spark]` and then import the new SparkCompare via `from datacompy.spare_core import SparkCompare`.

Closes #13
Closes #147

### DIFF
--- a/datacompy/spark_core.py
+++ b/datacompy/spark_core.py
@@ -23,11 +23,8 @@ two dataframes.
 
 import logging
 import os
-import sys
 from enum import Enum
-from itertools import chain
 
-import numpy as np
 import pandas as pd
 import pyspark.sql
 from ordered_set import OrderedSet
@@ -39,6 +36,8 @@ from pyspark.sql.functions import (
     col,
     isnan,
     lit,
+    trim,
+    upper,
     when,
 )
 
@@ -124,18 +123,20 @@ class SparkCompare(BaseCompare):
     rel_tol : float, optional
         Relative tolerance between two values.
     df1_name : str, optional
-        A string name for the first dataframe.  This allows the reporting to
+        A string name for the first dataframe. This allows the reporting to
         print out an actual name instead of "df1", and allows human users to
         more easily track the dataframes.
     df2_name : str, optional
-        A string name for the second dataframe
+        A string name for the second dataframe. This allows the reporting to
+        print out an actual name instead of "df2", and allows human users to
+        more easily track the dataframes.
     ignore_spaces : bool, optional
         Flag to strip whitespace (including newlines) from string columns (including any join
         columns)
     ignore_case : bool, optional
         Flag to ignore the case of string columns
     cast_column_names_lower: bool, optional
-        Boolean indicator that controls of column names will be cast into lower case
+        Boolean indicator that controls if column names will be cast into lower case
 
     Attributes
     ----------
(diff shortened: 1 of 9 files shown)


fdosani:
> @fdosani @dan-coates Hi, I tried out this branch, and some basic usage was okay, but got error: **init**() got an unexpected keyword argument 'column_mapping' the 'column_mapping' option for SparkCompare() appears to be unimplemented. Some other options are unimplemented also...will those be supported?

@runrig How are you calling it? I've left out some stuff like `column_mapping` from the original to line it up more like the Pandas version. Current init signature options [here](https://github.com/capitalone/datacompy/blob/spark-compare-refactor/datacompy/spark_core.py#L150-L162). I can add the additional stuff in after, but just really want to get the basic functionality down to start.

fdosani:
@dan-coates @theianrobertson pinging. Any thoughts on this?

dan-coates:
Hey @fdosani, sorry I haven't had much time to look at this (and generally have not been very current on my GitHub notifications). Definitely don't wait up for me on this, but I'll see if I have time later today to look at things.

datacompy/spark_core.py
```diff
@@ -0,0 +1,1023 @@
...
+            raise ValueError("{} must have all columns from join_columns".format(index))
+
+        if len(set(dataframe.columns)) < len(dataframe.columns):
+            raise ValueError("{} must have unique column names".format(index))
+
+        if (
+            dataframe.drop_duplicates(subset=self.join_columns).count()
+            < dataframe.count()
+        ):
+            self._any_dupes = True
```
dan-coates, line 227: On the Spark side I think we have to be more mindful about the amount of time that actions on the dataframes will take, as the datasets may be arbitrarily large. This will walk through all rows of each dataframe, dedup them, and count both the deduplicated and original datasets. But if duplicates are found, it will perform the entire deduplication again in `_compare`. That could be an expensive operation and we should avoid trying to do it twice. That was the motivation behind [this code](https://github.com/capitalone/datacompy/blob/37b612a59b96e2ba41f1f1e7a08ac8972309dc16/datacompy/sparkcompare.py#L192-L199).

dan-coates (commented):
Hey @fdosani, it's been too long since I wrote the original code to recall the exact intent without spending more time on this than I have right now, but I left a few comments that hopefully help. I think the biggest thing to think of on the Spark side is that it is possible for the Spark datasets to be much larger than their Pandas counterparts, so trying to minimize the number of operations that may have to be repeated, or otherwise reduce the expense of the comparison is a good idea. I can't say how well I did that originally but that should be the guiding principle on the Spark side.

theianrobertson (commented):
Overall I think it's heading in a decent direction.  My thinking way back was that we try to avoid duplicating code unnecessarily, so abstract out the spark- or pandas-specific functions and then call them in things like the `report` function that can be shared from the base class.

To Dan's point, I think it'd probably be a good idea to profile the new Spark class vs. old Spark class on some large-ish Spark dataframes, to make sure it's not slowing right down.  I imagine that wouldn't be part of the CI suite, but maybe a one-off before release.

fdosani:
@theianrobertson @dan-coates Thanks for the helpful review. Sorry I've been really busy with mid year PM, so I'm starting to implement some of your changes you suggested. I might try and poke around this weekend or next as I'll have some time. Heading to SciPy to maybe in the evening would work. 

Again thanks for the feedback! Hoping to connect with you again soon.

fdosani:
Closing in favor of #195

PR #170, opened 2022-10-26
Use less restrictive dependency version constraints.

kleschenko:
## Changes
We use `datacompy` as a library in the project using `poetry` to lock dependencies. And using the strict upper bounds for minor versions makes it hard to update our dependencies to include bug and security fixes. This PR changes the version constraints to be more permissive.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,9 +33,9 @@ zip_safe = False
 include_package_data = True
 packages = find:
 install_requires = 
-	pandas<=1.5.1,>=0.25.0
-	numpy<=1.23.4,>=1.11.3
-	ordered-set<=4.1.0,>=4.0.2
+	pandas<=1.6,>=0.25.0
+	numpy<=1.24,>=1.11.3
+	ordered-set<=5.0,>=4.0.2
 
 [options.package_data]
 * = templates/*


fdosani:
Hi @kleschenko thanks for the PR. Currently the 3 dependencies for `datacompy` are using the latest version. There shouldn't be any concerns in that regard. We also use [edgetest](https://github.com/capitalone/edgetest) (whcih I maintain) to help bump up version each week (assuming they pass unit testing). You can see an example [here](https://github.com/capitalone/datacompy/pull/168/files)

Is this for more future issues and trying to get a head of it? We should be covered with edgetest and maybe automation of released based on it. Appreciate your thoughts on the above or if there are other concerns? The changes you are suggesting would get written over by the edgetest action on Friday if merged.

kleschenko:
> Currently the 3 dependencies for datacompy are using the latest version.

They are using the latest version in the repository, but not in the [latest published](https://github.com/capitalone/datacompy/blob/v0.8.2/setup.cfg#L35) package, and that's where the main problem is. A new `datacompy` release [won't always be available](https://github.com/capitalone/datacompy/releases) immediately after the dependency bump, and some users could be stuck on the vulnerable (or even broken for them) version of the dependency. That's why it would be nice to have a room for updates without waiting for the new release, and it is considered a good practice nowadays to keep the library dependency constraints pretty relaxed (opposite to the application constraints where you want to have everything pinned to the exact versions), unless it is absolutely necessary to avoid the new dependency versions.

I'm not sure if that practice will work well for the `datacompy` library and would appreciate to hear your thoughts and concerns about it

fdosani:
Maybe I’m not following. The dev branch is up to date. So if the thought is folks want to install from there that’s currently available. If it’s helpful to cut a release right now I can. I’m just not sure the changes suggested are going to deliver a different outcome than what currently exists. As a maintainer I want to make sure new dependency releases don’t cause breaking changes which is possible and at a minimum pass the unit testing. Hence Edgetest.   On Wed, Oct 26, 2022 at 7:46 PM, Kostya Leschenko ***@***.***> wrote:  

Currently the 3 dependencies for datacompy are using the latest version.

They are using the latest version in the repository, but not in the latest published package, and that's where the main problem is. A new datacompy release won't always be available immediately after the dependency bump, and some users could be stuck on the vulnerable (or even broken for them) version of the dependency. That's why it would be nice to have a room for updates without waiting for the new release, and it is considered a good practice nowadays to keep the library dependency constraints pretty relaxed (opposite to the application constraints where you want to have everything pinned to the exact versions), unless it is absolutely necessary to avoid the new dependency versions.
I'm not sure if that practice will work well for the datacompy library and would appreciate to hear your thoughts and concerns about it

fdosani:
Feel free to ping me here, but I'm going to close this for now since I think given our usage of `edgetest` this might not be appropriate. We can automated releases on a more frequent basis to overcome some concerns here.

Issue #177, opened 2022-12-15
all_mismatch method raise exception if row count is different between baseline dataframe with compared dataframe 


Sunny08012012:
if baseline dataframe rows is different from compared dataframe rows, when using all_mismatch, will raise below exception 
    **return less_equal(abs(x-y), atol + rtol * abs(y))
ValueError: operands could not be broadcast together with shapes (7,) (5,)** 

example is:
data11 = """acct_id,dollar_amt,name,float_fld
10000001234,123.45,George Maharis,14530.1555
10000001235,0.45,Michael Bluth,1
10000001236,1345,George Bluth,
10000001237,123456,Bob Loblaw,345.12
10000001239,1.05,Lucille Bluth,
10000001240,11.05,Lucille Bluth,
10000001241,1111.05,Lucille Bluth,
"""
data2 = """acct_id,dollar_amt,name,float_fld
10000001234,123.4,George Michael Bluth,14530.155
10000001235,0.45,Michael Bluth,1
10000001236,1345,George Bluth,
10000001237,123456,Robert Loblaw,345.12
10000001238,1.05,Loose Seal Bluth,111
"""

df1 = pd.read_csv(StringIO(data11))
df2 = pd.read_csv(StringIO(data2))

compare = datacompy.Compare(
    df1,
    df2,
    join_columns='acct_id',  #You can also specify a list of columns
    abs_tol=0, #Optional, defaults to 0
    rel_tol=0, #Optional, defaults to 0
    df1_name='baseline', #Optional, defaults to 'df1'
    df2_name='tested', #Optional, defaults to 'df2'
    ignore_spaces=True,
    ignore_case=True
    )

df_mismatch=compare.all_mismatch(ignore_matching_cols=True)
print(df_mismatch)

fdosani:
Seems like a bug from #166 . Working on a fix.

fdosani:
@Sunny08012012 I've made a fix here: https://github.com/capitalone/datacompy/compare/develop...all-mismatch-bug-fix
Would you mind installing from this branch and testing again. This should resolve the exception you were facing.

PR #178, opened 2022-12-15
Bug fix for #177

fdosani:
Seems like a minor bug where if the columns being compared had different count of rows it would hit an exception on the compare. Switching to `intersect_rows` from` df1/df2` should fix this as we want the intersect.


fdosani:
@ak-gupta if you have time

ak-gupta (approved):

PR #185, opened 2023-01-29
Update code with f-strings 

pronoym99:
Update code with `f-strings` except those with SQL statements and separate format specs


pronoym99:
I have kept these two cases for later evaluation on a case-by-case basis.

- SQL statements invite injection vulnerabilities in all but one type of formatting methods i.e. `%` - formatting.
- The usage of specific `format` specs in multiple scenarios will need another round of evaluation and review.

fdosani (approved):

PR #195, opened 2023-05-03
SparkCompare refactor

### LINKED ISSUE #13: Align the APIs between Compare and SparkCompare

### LINKED ISSUE #147: Use "<=>" operator to join spark dataframes

fdosani:
Revisiting the refactor from #148 
This is taking a different approach in that we are using the Pandas on Spark API vs just vanilla PySpark.

Following the advice from the original discussion and make a `ABC` based on the pandas versions, and then implement the same set of items in Spark. There are some differences which I should note:

- No indexing like pandas has. Joining is done on `join_columns` like the old SparkCompare.
- Dups are preserved like Pandas
- In the new SparkCompare we are able to do a null-safe joins now on the join columns. This was requested by @runrig in #147
- Spark 3.4+ will only be supported moving forward.

Will start this off as a draft just to make sure folks are aligned. @runrig, if you have time to clone this branch and check out the new refactor I'd appreciate it. It should give you what you were looking for in `<=>` joins. You should be able to `pip install -e .[dev,spark]` and then import the new SparkCompare via `from datacompy.spare_core import SparkCompare`.

Closes #148 
Closes #13
Closes #147


fdosani:
@NikhilJArora @ak-gupta if you have some time to help review this, I'd really appreciate it.

fdosani:
@goodwanghan Just for visibility. This PR might be moot given what we discussed about [fugue](https://github.com/fugue-project/fugue) integration.

fdosani:
To do:
- reinsert back old compare and do some form of deprecation to enable some compatibility
- rename the new `SparkCompare` to `PandasSparkCompare`.
- Docs for Fugue and outlining usage

fdosani:
Going to close this PR for now. Its been sitting, and not sure Koalas is the way to really go with this. FYI @jdawang

PR #201, opened 2023-05-08
Add Fugue support (Phase 1)

goodwanghan:
[Fugue](https://github.com/fugue-project/fugue) is an abstraction for distributed and local computing frameworks such as Spark Dask, Ray, Duckdb and Polars. For datacompy, fugue can elegantly scale the core Compare class to different distributed backends.

Is Phase 1, we only implemented `is_match` function. In Phase 2, we will enable report.

Notice, `is_match` will compare unordered data, meaning that, `df1`, `df2` without the same order can still match. In distributed systems, the concept of order doesn't natively exist, so that is why `is_match` doesn't require orders.

Here are a few examples to use is_match

```python
import fugue.api as fa
from datacompy import is_match

is_match(pdf1, pdf2, join_columns="a")  # defaults to Compare class
is_match(pdf1, pdf2, join_columns="a", parallelism=1)  # force to use Fugue, but the backend is still pandas

is_match(spark_df1, spark_df2, join_columns="a") # compare spark dataframes using spark
is_match(spark_df1, pdf2, join_columns="a") # compare spark dataframe with pandas dataframe using spark
is_match(pdf1, spark_df2, join_columns="a") # compare spark dataframe with pandas dataframe using spark

with fa.engine_context(spark_session):
    is_match(pdf1, pdf2, join_columns="a")  # force to use spark to compare dataframes

is_match(ray_df1, ray_df2, join_columns="a") # compare ray dataset using ray
is_match(ray_df1, pdf2, join_columns="a") # compare ray dataset with pandas dataframe using ray
is_match(pdf1, ray_df2, join_columns="a") # compare ray dataset with pandas dataframe using ray
```

### DIFF
--- a/datacompy/fuguecompare.py
+++ b/datacompy/fuguecompare.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2020 Capital One Services, LLC
+# Copyright 2023 Capital One Services, LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -77,6 +77,14 @@ def is_match(
         Flag to ignore the case of string columns
     cast_column_names_lower: bool, optional
         Boolean indicator that controls of column names will be cast into lower case
+    parallelism: int, optional
+        An integer representing the amount of parallelism. Entering a value for this
+        will force to use of Fugue over just vanilla Pandas
+
+    Returns
+    -------
+    bool
+        Returns boolean as to if the DataFrames match.
     """
     if (
         isinstance(df1, pd.DataFrame)
--- a/datacompy/fuguecompare.py
+++ b/datacompy/fuguecompare.py
@@ -44,6 +44,7 @@ def is_match(
     ignore_case: bool = False,
     cast_column_names_lower: bool = True,
     parallelism: Optional[int] = None,
+    strict_schema: bool = False,
 ) -> bool:
     """Check whether two dataframes match.
 
@@ -80,6 +81,8 @@ def is_match(
     parallelism: int, optional
         An integer representing the amount of parallelism. Entering a value for this
         will force to use of Fugue over just vanilla Pandas
+    strict_schema: bool, optional
+        The schema must match exactly if set to ``True``. This includes the names and types. Allows for a fast fail.
 
     Returns
     -------
@@ -113,8 +116,6 @@ def is_match(
         hash_cols = [join_columns]
     else:
         hash_cols = join_columns
-    # ensure all cols exist
-    hash_cols = tdf1.schema.extract(hash_cols).names
 
     if cast_column_names_lower:
         tdf1 = tdf1.rename(
@@ -125,7 +126,12 @@ def is_match(
         )
         hash_cols = [col.lower() for col in hash_cols]
 
-    if tdf1.schema != tdf2.schema:
+    if strict_schema:
+        if tdf1.schema != tdf2.schema:
+            return False
+
+    # check that hash columns exist
+    if hash_cols not in tdf1.schema or hash_cols not in tdf2.schema:
         return False
 
     all_cols = tdf1.schema.names
--- a/datacompy/fuguecompare.py
+++ b/datacompy/fuguecompare.py
@@ -1,0 +1,189 @@
+#
+# Copyright 2020 Capital One Services, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Compare two DataFrames that are supported by Fugue
+"""
+
+import logging
+import pickle
+from typing import Any, Dict, Iterable, List, Optional, Union
+
+import fugue.api as fa
+import pandas as pd
+import pyarrow as pa
+from fugue import AnyDataFrame
+
+from .core import Compare
+
+LOG = logging.getLogger(__name__)
+HASH_COL = "__datacompy__hash__"
+
+
+def is_match(
+    df1: AnyDataFrame,
+    df2: AnyDataFrame,
+    join_columns: Union[str, List[str]],
+    abs_tol: float = 0,
+    rel_tol: float = 0,
+    df1_name: str = "df1",
+    df2_name: str = "df2",
+    ignore_spaces: bool = False,
+    ignore_case: bool = False,
+    cast_column_names_lower: bool = True,
+    parallelism: Optional[int] = None,
+) -> bool:
+    """Check whether two dataframes match.
+
+    Both df1 and df2 should be dataframes containing all of the join_columns,
+    with unique column names. Differences between values are compared to
+    abs_tol + rel_tol * abs(df2['value']).
+
+    Parameters
+    ----------
+    df1 : ``AnyDataFrame``
+        First dataframe to check
+    df2 : ``AnyDataFrame``
+        Second dataframe to check
+    join_columns : list or str, optional
+        Column(s) to join dataframes on.  If a string is passed in, that one
+        column will be used.
+    abs_tol : float, optional
+        Absolute tolerance between two values.
+    rel_tol : float, optional
+        Relative tolerance between two values.
+    df1_name : str, optional
+        A string name for the first dataframe.  This allows the reporting to
+        print out an actual name instead of "df1", and allows human users to
+        more easily track the dataframes.
+    df2_name : str, optional
+        A string name for the second dataframe
+    ignore_spaces : bool, optional
+        Flag to strip whitespace (including newlines) from string columns (including any join
+        columns)
+    ignore_case : bool, optional
+        Flag to ignore the case of string columns
+    cast_column_names_lower: bool, optional
+        Boolean indicator that controls of column names will be cast into lower case
+    """
+    if (
+        isinstance(df1, pd.DataFrame)
+        and isinstance(df2, pd.DataFrame)
+        and parallelism is None  # user did not specify parallelism
+        and fa.get_current_parallelism() == 1  # currently on a local execution engine
+    ):
+        comp = Compare(
+            df1=df1,
+            df2=df2,
+            join_columns=join_columns,
+            abs_tol=abs_tol,
+            rel_tol=rel_tol,
+            df1_name=df1_name,
+            df2_name=df2_name,
+            ignore_spaces=ignore_spaces,
+            ignore_case=ignore_case,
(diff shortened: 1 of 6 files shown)


goodwanghan:
@kvnkho @fdosani

fdosani:
@goodwanghan 

Just poking around with the code base. If I do something like the following. Basically comparing 2 dataframes which are just sorted backwards it returns different results which from reading the code makes sense.

```
import pandas as pd
import numpy as np
from datacompy import is_match


df1 = pd.DataFrame(np.random.randint(0, 100, size=(10000, 2)), columns=["b", "c"])
df1.reset_index(inplace=True)
df1.columns = ["a", "b", "c"]
df2 = df1.copy()
df2["b"] = df2["b"] + 0.1
df2.sort_index(ascending=False, inplace=True)

In [55]: is_match(df1, df2, join_columns="a", parallelism=1, abs_tol=0.2)
Out[55]: False

In [56]: is_match(df1, df2, join_columns="a", abs_tol=0.2)
Out[56]: True
```

Fully admit being new to Fugue I might not totally understand the implementation so please excuse my ignorance or dumb questions.

So the splitting up into groups happens [here](https://github.com/capitalone/datacompy/pull/201/files#diff-b9e013e80e354707f45a01e030cfb8a25ed6fcc8c8a668027e976389af9f294bR142-R143). So lets say I have 3 "buckets" those 3 may or may not have the corresponding row from the other dataframe.

Based on what @kvnkho mentioned at PyData these buckets are what is run on say Spark correct in isolation?
I'm wondering if there is a way to join things before bucketing (this might not be simple or scaleable) hence we really want to join things to compare vs just comparing rows.

goodwanghan:
> > > is_match(df1, df2, join_columns="a", parallelism=1, abs_tol=0.2)
> > 
> > 
> > Ah, this is because they don't match on schema. df1.b is integer df2.b is float. Fugue has more strict schema comparison. But of course, if you want to loose this, we can hand special logic to handle it.
> > if you add
> > ```
> > df1["b"]=df1["b"].astype(float)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > you will see the compare succeeds
> 
> That is great. Ok so maybe my understanding was off here. The [_comp](https://github.com/capitalone/datacompy/pull/201/files#diff-b9e013e80e354707f45a01e030cfb8a25ed6fcc8c8a668027e976389af9f294bR161) which happens does happen on the entire dataset then correct? I thought maybe pieces are being segmented off and that if things were not ordered properly there would be a chance the join wouldn't take place.

The `_comp` function happens on each partition that will be handled by a worker.
Each partition will contain multiple groups of data defined by joined_columns

For example

df1

a b
x 1
x 2
y 1
y 2
z 1

df2

a b
x 1
x 2
y 1
y 2
z 1
z 2

if join_columns is a, then the data could be partitioned as group 1 (containing x y), and group 2(containing z)
Then on the remote worker, for each partition, the subdataframes will be reconstructed and compared using `_comp`

goodwanghan:
> @goodwanghan
> 
> Just poking around with the code base. If I do something like the following. Basically comparing 2 dataframes which are just sorted backwards it returns different results which from reading the code makes sense.
> 
> ```
> import pandas as pd
> import numpy as np
> from datacompy import is_match
> 
> 
> df1 = pd.DataFrame(np.random.randint(0, 100, size=(10000, 2)), columns=["b", "c"])
> df1.reset_index(inplace=True)
> df1.columns = ["a", "b", "c"]
> df2 = df1.copy()
> df2["b"] = df2["b"] + 0.1
> df2.sort_index(ascending=False, inplace=True)
> 
> In [55]: is_match(df1, df2, join_columns="a", parallelism=1, abs_tol=0.2)
> Out[55]: False
> 
> In [56]: is_match(df1, df2, join_columns="a", abs_tol=0.2)
> Out[56]: True
> ```
> 
> Fully admit being new to Fugue I might not totally understand the implementation so please excuse my ignorance or dumb questions.
> 
> So the splitting up into groups happens [here](https://github.com/capitalone/datacompy/pull/201/files#diff-b9e013e80e354707f45a01e030cfb8a25ed6fcc8c8a668027e976389af9f294bR142-R143). So lets say I have 3 "buckets" those 3 may or may not have the corresponding row from the other dataframe.
> 
> Based on what @kvnkho mentioned at PyData these buckets are what is run on say Spark correct in isolation? I'm wondering if there is a way to join things before bucketing (this might not be simple or scaleable) hence we really want to join things to compare vs just comparing rows.

In this case, `bucket=2`

You asked about join, we didn't directly use join here, instead we use map->union->groupmap, which is similar to join.

fdosani:
> You asked about join, we didn't directly use join here, instead we use map->union->groupmap, which is similar to join.

Got it! Yup this is making sense now. Thanks for the explanation. 

Are you OK if I push a couple of small tweaks? Nothing logic wise. This all makes sense now. 
More just naming and organization.

goodwanghan:
> > You asked about join, we didn't directly use join here, instead we use map->union->groupmap, which is similar to join.
> 
> Got it! Yup this is making sense now. Thanks for the explanation.
> 
> Are you OK if I push a couple of small tweaks? Nothing logic wise. This all makes sense now. More just naming, organization, and docs.
> 
> I guess maybe the other thing we should discuss is the strict vs loose comparison we discussed [above](https://github.com/capitalone/datacompy/pull/201#issuecomment-1539006805). In the core.py it happens [here](https://github.com/capitalone/datacompy/blob/dd1f61ca9831246469225f941d8e22fc6f03eb23/datacompy/core.py#L695) in `columns_equal`

Please feel free to make changes. Yeah we can also implement `columns_equal`, that is straightforward.

fdosani:
> > is_match(df1, df2, join_columns="a", parallelism=1, abs_tol=0.2)
> 
> Ah, this is because they don't match on schema. df1.b is integer df2.b is float. Fugue has more strict schema comparison. But of course, if you want to loose this, we can hand special logic to handle it.
> 
> if you add
> 
> ```
> df1["b"]=df1["b"].astype(float)
> ```
> 
> you will see the compare succeeds

Sorry just one thing about this strict schema you noted. Shouldn't it work as is, without the `astype`? 
When `Compare` is called it should run the internal logic in `columns_equal`. As it does with just native pandas dataframes. I guess I'm missing what `fugue` is doing in the background with the schema and types.

datacompy/fuguecompare.py
```diff
@@ -0,0 +1,189 @@
...
+    if cast_column_names_lower:
+        tdf1 = tdf1.rename(
+            {col: col.lower() for col in tdf1.schema.names if col != col.lower()}
+        )
+        tdf2 = tdf2.rename(
+            {col: col.lower() for col in tdf2.schema.names if col != col.lower()}
+        )
+        hash_cols = [col.lower() for col in hash_cols]
+
+    if tdf1.schema != tdf2.schema:
```
fdosani, line 120: I think this can be removed I'm not mistaken. The internals of Compare will try casting and various check to see if will work.

goodwanghan: This is to fail fast, it can fail without any real compute. For big data this can save a lot of time and money. But yeah, shall we make this optional?

fdosani: :+1:  I think the optional makes sense. Something like `strict_schema`?

But in the case where the "types" are different (but names match) then I think we should check to make sure we are checking. Like how the pandas compare behaves.

fdosani: Maybe something along the lines of: 

```python
    if strict_schema:
        if tdf1.schema != tdf2.schema:
            return False

    # check that hash columns exist
    if hash_cols not in tdf1.schema or hash_cols not in tdf2.schema:
        return False
```

datacompy/fuguecompare.py
```diff
@@ -0,0 +1,189 @@
...
+            for name in df.columns:
+                col = df[name]
+                if name in str_cols:
+                    if ignore_spaces:
+                        col = col.str.strip()
+                    if ignore_case:
+                        col = col.str.lower()
+                cols[name] = col
+            data = pd.DataFrame(cols)
+            gp = pd.util.hash_pandas_object(df[hash_cols], index=False).mod(bucket)
```
fdosani, line 141: Should `hash_cols` and `bucket` be arguments passed into `_serialize`? It seems they are being picked up from the scope above. Feel free to ignore if this doesn't make sense.

datacompy/fuguecompare.py
```diff
@@ -0,0 +1,189 @@
...
+            schema="key:int,left:bool,data:binary",
+            params=dict(left=False),
+        ),
+        distinct=False,
+    )
+
+    def _comp(df: List[Dict[str, Any]]) -> List[List[Any]]:
+        df1 = (
+            pd.concat([pickle.loads(r["data"]) for r in df if r["left"]])
+            .sort_values(all_cols)
```
fdosani, line 164: Similar question as above. `all_cols`, `join_columns`, `abs_tol`, etc.. are all being picked up from outside of `_comp`. Might be OK but just thought I'd ask.

datacompy/fuguecompare.py
goodwanghan: Yes, that is fine. This is why I defined function inside function in order to achieve this. It makes code simpler.

datacompy/fuguecompare.py
goodwanghan: Same, this is why I use the embedded function.

datacompy/fuguecompare.py
goodwanghan: Yes I agree with strict schema. But for missing cols, I ensured that here:
```
hash_cols = tdf1.schema.extract(hash_cols).names
```
If the join cols user provided are not found in the schema, I think it should raise an exception. This is not data issue, the user breaks the contract.

goodwanghan:
I did a test on 100gb data. It can finish in 7 min with 256 cpus.

The test is that I loaded two spark dataframes from the same file. `is_match` returns true, but SparkCompare reported difference.

datacompy/fuguecompare.py
fdosani: Agreed! We would also need to check `tdf2` to ensure it doesn't throw an exception here too.

goodwanghan: So should I make the change, or you can cover that?

datacompy/fuguecompare.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+        hash_cols = [col.lower() for col in hash_cols]
+
+    if strict_schema:
+        if tdf1.schema != tdf2.schema:
+            return False
+
+    # check that hash columns exist
+    if hash_cols not in tdf1.schema or hash_cols not in tdf2.schema:
+        return False
```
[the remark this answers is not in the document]
goodwanghan: Yes, this looks good to me!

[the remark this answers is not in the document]
fdosani: Awesome. I'll merge this in shortly. Thanks again for your help here!
Just want to get to a interim release for some other `SparkCompare` items. Shouldn't be long.

fdosani (approved):

datacompy/fuguecompare.py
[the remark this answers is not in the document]
goodwanghan: Thanks a lot! looking forward to it!

Issue #193, opened 2023-04-14
Error when comparing pandas string type with NAs


robne1982:
There appears to be an issue comparing series of pandas string types when that series contains NAs.

Minimal example below, along with the error message.

Partial workaround is to set the `sample_count` to 0 in the `.report()` call (e.g. below, `print(cmp_str.report(sample_count=0))`). Alternatively casting back to object type also works

I'll try to create a PR to fix.

-----------------------------

```
import datacompy
import pandas as pd

# create a pair of dfs with a simple NA vs value mismatch
# one using objects, one using pandas string type
dfa_obj = pd.DataFrame({"a": ["A", pd.NA]})
dfa_str = dfa_obj.copy()
dfa_str["a"] = dfa_str["a"].astype("string")

dfb_obj = pd.DataFrame({"a": ["A", "A"]})
dfb_str = dfb_obj.copy()
dfb_str["a"] = dfb_str["a"].astype("string")

# runs ok
cmp_obj = datacompy.Compare(df1=dfa_obj, df2=dfb_obj, on_index=True)
print(cmp_obj.report())

# errors
cmp_str = datacompy.Compare(df1=dfa_str, df2=dfb_str, on_index=True)
print(cmp_str.report())
```

the final line returns

```
Traceback (most recent call last):
  File "<REMOVED>\datacompy-play\main.py", line 20, in <module>
    print(cmp_str.report())
  File "<REMOVED>\\datacompy-play\lib\site-packages\datacompy\core.py", line 624, in report
    self.sample_mismatch(
  File"<REMOVED>\\datacompy-play\lib\site-packages\datacompy\core.py", line 480, in sample_mismatch
    sample = self.intersect_rows[~col_match].sample(sample_count)
  File ""<REMOVED>\\datacompy-play\lib\site-packages\pandas\core\generic.py", line 5773, in sample
    sampled_indices = sample.sample(obj_len, size, replace, weights, rs)
  File ""<REMOVED>\\datacompy-play\lib\site-packages\pandas\core\sample.py", line 150, in sample
    return random_state.choice(obj_len, size=size, replace=replace, p=weights).astype(
  File "mtrand.pyx", line 928, in numpy.random.mtrand.RandomState.choice
ValueError: a must be greater than 0 unless no samples are taken
```

fdosani:
Thanks @rnobleeddy Appreciate the help here.

PR #207, opened 2023-05-27
Add support for stings with pd.NA

robne1982:
UPDATE - will need to fix the spark tests - didn't do this locally.

WIP - if this approach is acceptable, I can update the current tests that use strings to run with series of object and string type.

The PR:

- updates core.py to prevent comparisons with `pd.NA` from returning `pd.NA`. The downside of this is that I believe any kinds of missing will match, so it won't be possible to distinguish `pd.NA` and `NaN`, for example.
- adds a test to check behaviour for strings is as expected (and checks it works for objects too)
- updates one of the existing tests with strings to use both dtypes

TODO:

- understand if this approach is acceptable for the package owners
- update the other existing tests with strings to use both object and string dtypes
- add some tests to cover the behaviour for matching different kinds of missing data, e.g. pd.NA and nan

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -760,7 +760,7 @@ def columns_equal(
                     compare = compare_string_and_date_columns(col_1, col_2)
                 else:
                     compare = pd.Series(
-                        (col_1 == col_2) | (col_1.isnull() & col_2.isnull())
+                        ((col_1 == col_2).combine_first(col_1.isna() & col_2.isna())) | (col_1.isnull() & col_2.isnull())
                     )
             except:
                 # Blanket exception should just return all False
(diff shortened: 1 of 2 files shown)


fdosani:
@rnobleeddy Thanks for the PR. I'm trying to dive into understanding the larger problem. Sorry been distracted of late with other things on the go, so apologies for the delay.

With `pd.NA` would it make more sense to convert the missing into something else (not pd.NA, but some text) before processing? Based on the Pandas docs pd.NA is [experimental](https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html?highlight=na#experimental-na-scalar-to-denote-missing-values) and thinking about it a bit more I'm not sure what the implications are of adding in this functionality. This could potentially breaking things for other cases or cause weird behavior in future Pandas versions.

Looking back into the issue (https://github.com/capitalone/datacompy/issues/193). Are you getting `pd.NA` when reading in a file? I guess this is specific only to the "string" type vs object.

PR #217, opened 2023-06-21
adding in unq_columns and intersect_columns for Fugue

fdosani:
Adding in `unq_columns` for the Fugue module. Equivalent of the:
- df1_unq_columns
- and df2_unq_columns

Fairly simple function which mirrors the `core` version. Main difference is that I used `as_fugue_df` to conver the inputs to ensure uniformity and compatability between different frameworks. @goodwanghan does this make sense? Just want to confirm my thought process here.

### DIFF
--- a/datacompy/fugue.py
+++ b/datacompy/fugue.py
@@ -50,9 +50,9 @@ def unq_columns(df1: AnyDataFrame, df2: AnyDataFrame):
     OrderedSet
         Set of columns that are unique to df1
     """
-    tdf1 = fa.as_fugue_df(df1)
-    tdf2 = fa.as_fugue_df(df2)
-    return OrderedSet(tdf1.columns) - OrderedSet(tdf2.columns)
+    col1 = fa.get_column_names(df1)
+    col2 = fa.get_column_names(df2)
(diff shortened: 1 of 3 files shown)


datacompy/fugue.py
```diff
@@ -26,13 +26,35 @@
...
+        Second dataframe to check
+
+    Returns
+    -------
+    OrderedSet
+        Set of columns that are unique to df1
+    """
+    tdf1 = fa.as_fugue_df(df1)
+    tdf2 = fa.as_fugue_df(df2)
+    return OrderedSet(tdf1.columns) - OrderedSet(tdf2.columns)
```
goodwanghan, line 55: actually, we shouldn't convert them to fugue df in this case.
https://fugue.readthedocs.io/en/latest/top_api.html#fugue.api.get_column_names

```python
columns1 = fa.get_column_names(df1)
...
```

This can bypass any unnecessary data conversion.

fdosani: Perfect!

fdosani:
Including `intersect_columns` as well since it is similar to `unq_columns`

goodwanghan (approved):
Looks great to me

PR #226, opened 2023-07-11
removing requirements.txt

fdosani:
We don't really need a specific `requirements.txt` since developers can just `pip install -e .[dev]`


fdosani:
@NikhilJArora @ak-gupta ping 😄

NikhilJArora:
I think we need to remove the ref to `requirements.txt` in this workflow with this change. https://github.com/capitalone/datacompy/blob/1de2f37f75064422be5f82cd8f35a0859f0fb8a0/.github/workflows/publish-package.yml#L23

fdosani:
Closing in favour of #229

PR #236, opened 2023-09-18
fix for logging intersect rows number

BrainLogic:
fix for logging `intersect rows` number


fdosani (approved):
Thanks for finding and fixing this.

Issue #214, opened 2023-06-06
Fugue support for extra helper functions from core


fdosani:
Currently there are some helper functions as part of the core Pandas code which I think are generally very helpful.
We need to spend some time exploring those and seeing which ones can be mirrored/included via the Fugue implementation.

This is just a list of most of the functions. Not all will make sense to move over. But we should investigate which ones make sense to:

- df1_unq_columns
- df2_unq_columns
- intersect_columns
- all_columns_match
- all_rows_overlap
- count_matching_rows
- intersect_rows_match
- matches
- subset
- sample_mismatch
- all_mismatch
- columns_equal
- compare_string_and_date_columns
- get_merged_columns
- temp_column_name
- calculate_max_diff
- generate_id_within_group

fdosani:
@goodwanghan @kvnkho FYI, no pressure to contribute, but something in our backlog I'm thinking to ensure full parity in terms of function etc.

goodwanghan:
Sounds good, let's chat about it

fdosani:
Closing this issue as it is stale.

PR #244, opened 2023-11-10
adding all_rows_overlap to fugue

fdosani:
Ref: https://github.com/capitalone/datacompy/issues/214
Support for all_rows_overlap with Fugue

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -18,6 +18,7 @@ __version__ = "0.10.5"
 from datacompy.core import *
 from datacompy.fugue import (
     all_columns_match,
+    all_rows_overlap,
     intersect_columns,
     is_match,
     report,
--- a/datacompy/fugue.py
+++ b/datacompy/fugue.py
@@ -195,6 +195,102 @@ def is_match(
     return all(matches)
 
 
+def all_rows_overlap(
+    df1: AnyDataFrame,
+    df2: AnyDataFrame,
+    join_columns: Union[str, List[str]],
+    abs_tol: float = 0,
+    rel_tol: float = 0,
+    df1_name: str = "df1",
+    df2_name: str = "df2",
+    ignore_spaces: bool = False,
+    ignore_case: bool = False,
+    cast_column_names_lower: bool = True,
+    parallelism: Optional[int] = None,
+    strict_schema: bool = False,
+) -> bool:
+    """Check if the rows are all present in both dataframes
+
+    Parameters
+    ----------
+    df1 : ``AnyDataFrame``
+        First dataframe to check
+    df2 : ``AnyDataFrame``
+        Second dataframe to check
+    join_columns : list or str, optional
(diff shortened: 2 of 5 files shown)


fdosani:
@goodwanghan FYI. Had some time, so adding in some more functionality to the Fugue side.

goodwanghan:
Wow nice

fdosani:
> Wow nice

Well this sucks. Seems like the last release of fugue is breaking some of the tests :(

goodwanghan:
Oh what? I will take a look

goodwanghan:
I can reproduce this issue, will take a look today

goodwanghan:
So after some initial investigation, it is related with pandas. Fugue 0.8.7 has a big update that is it relies more on pandas extension d types. It is an interesting case why the current datacompy logic breaks with this change. I will investigate more.

fdosani:
> So after some initial investigation, it is related with pandas. Fugue 0.8.7 has a big update that is it relies more on pandas extension d types. It is an interesting case why the current datacompy logic breaks with this change. I will investigate more.

Ahh interesting. Thanks for helping out here and taking a peek, really appreciate it.

goodwanghan:
no problem, actually think this is a very good case, i have found an example
You know pandas will use object type for string columns, but in new pandas, it can be StringDType, and it can be an arrow type (it is a mess I know). So in this case we want to treat them as the same type? In the current logic, we think they are different. That is why we saw those failed tests.

fdosani:
Gotcha that makes sense. Yeah strings should in theory be the same in that example.   On Fri, Nov 10, 2023 at 1:34 PM, Han Wang ***@***.***> wrote:  
no problem, actually think this is a very good case, i have found an example
You know pandas will use object type for string columns, but in new pandas, it can be StringDType, and it can be an arrow type (it is a mess I know). So in this case we want to treat them as the same type? In the current logic, we think they are different. That is why we saw those failed tests.

fdosani:
> no problem, actually think this is a very good case, i have found an example
> 
> You know pandas will use object type for string columns, but in new pandas, it can be StringDType, and it can be an arrow type (it is a mess I know). So in this case we want to treat them as the same type? In the current logic, we think they are different. That is why we saw those failed tests.
> 
> 
> 
> To generalize, datacompy needs to consider the comparison between original type and extension dtypes, do we think they are the same?
> 
> 
> 
> Examples are:
> 
> 
> 
> object vs StringDType
> 
> int vs Int
> 
> bool vs Bool
> 
> 
> 
> 
> 
> 

Yup those should be considered the same type IMHO.

jdawang:
We should probably list the added functionality for fugue in the docs. Right now only `is_match` and `report` are mentioned.

fdosani:
> We should probably list the added functionality for fugue in the docs. Right now only `is_match` and `report` are mentioned.

Good call! Will add this in a bit.

fdosani:
@jdawang updated now

jdawang (approved):

PR #249, opened 2023-12-07
✨ Add mypy for static type checking and type annotations

aguiddir:
### ✨ Add mypy for static type checking and type annotations

This pull request introduces `mypy` for static type checking and incorporates type annotations across the `datacompy` project. The integration encompasses installing `mypy` as a development dependency, configuring `pyproject.toml`, and systematically adding type hints to functions, methods, and variables throughout the codebase.

## Changes Made

- Installed `mypy` as a qa dependency to enable static type checking.
- Added necessary configuration to `pyproject.toml` for `mypy`.
- Annotated functions, methods, and variables with type hints to improve code maintainability and readability.

## Purpose

The primary goal of this enhancement is to identify potential type-related errors early in the development phase, fostering a more resilient codebase. Additionally, the inclusion of type annotations aims to enhance code comprehension for both existing and future contributors.

## Testing

- Ensure `mypy` is installed (if not, install it via `pip install mypy`).
- Run `mypy` on the project directory to perform static type checking (`mypy path/to/datacompy`) and verify absence of type-related errors or warnings.

## TODO

- Add `mypy` to CI

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -20,9 +20,9 @@ Originally this package was meant to provide similar functionality to
 PROC COMPARE in SAS - i.e. human-readable reporting on the difference between
 two dataframes.
 """
-
 import logging
 import os
+from typing import cast, Any
 
 import numpy as np
 import pandas as pd
@@ -79,18 +79,18 @@ class Compare:
 
     def __init__(
         self,
-        df1,
-        df2,
-        join_columns=None,
-        on_index=False,
-        abs_tol=0,
-        rel_tol=0,
-        df1_name="df1",
-        df2_name="df2",
-        ignore_spaces=False,
-        ignore_case=False,
-        cast_column_names_lower=True,
-    ):
+        df1: pd.DataFrame,
+        df2: pd.DataFrame,
+        join_columns: list[str] | str | None = None,
+        on_index: bool = False,
+        abs_tol: float = 0,
+        rel_tol: float = 0,
+        df1_name: str = "df1",
+        df2_name: str = "df2",
+        ignore_spaces: bool = False,
+        ignore_case: bool = False,
+        cast_column_names_lower: bool = True,
+    ) -> None:
         self.cast_column_names_lower = cast_column_names_lower
         if on_index and join_columns is not None:
             raise Exception("Only provide on_index or join_columns")
@@ -107,11 +107,11 @@ class Compare:
         else:
             self.join_columns = [
                 str(col).lower() if self.cast_column_names_lower else str(col)
-                for col in join_columns
+                for col in cast(list[str], join_columns)
             ]
             self.on_index = False
 
-        self._any_dupes = False
+        self._any_dupes: bool = False
         self.df1 = df1
         self.df2 = df2
         self.df1_name = df1_name
@@ -120,16 +120,18 @@ class Compare:
         self.rel_tol = rel_tol
         self.ignore_spaces = ignore_spaces
         self.ignore_case = ignore_case
-        self.df1_unq_rows = self.df2_unq_rows = self.intersect_rows = None
-        self.column_stats = []
-        self._compare(ignore_spaces, ignore_case)
+        self.df1_unq_rows: pd.DataFrame
+        self.df2_unq_rows: pd.DataFrame
+        self.intersect_rows: pd.DataFrame
+        self.column_stats: list[dict[str, Any]] = []
+        self._compare(ignore_spaces=ignore_spaces, ignore_case=ignore_case)
 
     @property
-    def df1(self):
+    def df1(self) -> pd.DataFrame:
         return self._df1
 
     @df1.setter
-    def df1(self, df1):
+    def df1(self, df1: pd.DataFrame) -> None:
         """Check that it is a dataframe and has the join columns"""
         self._df1 = df1
         self._validate_dataframe(
@@ -137,18 +139,20 @@ class Compare:
         )
 
     @property
-    def df2(self):
+    def df2(self) -> pd.DataFrame:
         return self._df2
 
     @df2.setter
-    def df2(self, df2):
+    def df2(self, df2: pd.DataFrame) -> None:
         """Check that it is a dataframe and has the join columns"""
         self._df2 = df2
         self._validate_dataframe(
             "df2", cast_column_names_lower=self.cast_column_names_lower
         )
 
-    def _validate_dataframe(self, index, cast_column_names_lower=True):
+    def _validate_dataframe(
+        self, index: str, cast_column_names_lower: bool = True
+    ) -> None:
         """Check that it is a dataframe and has the join columns
 
         Parameters
@@ -163,9 +167,11 @@ class Compare:
             raise TypeError(f"{index} must be a pandas DataFrame")
 
         if cast_column_names_lower:
-            dataframe.columns = [str(col).lower() for col in dataframe.columns]
+            dataframe.columns = pd.Index(
+                [str(col).lower() for col in dataframe.columns]
+            )
         else:
-            dataframe.columns = [str(col) for col in dataframe.columns]
+            dataframe.columns = pd.Index([str(col) for col in dataframe.columns])
         # Check if join_columns are present in the dataframe
         if not set(self.join_columns).issubset(set(dataframe.columns)):
             raise ValueError(f"{index} must have all columns from join_columns")
@@ -182,7 +188,7 @@ class Compare:
             ):
                 self._any_dupes = True
 
-    def _compare(self, ignore_spaces, ignore_case):
+    def _compare(self, ignore_spaces: bool, ignore_case: bool) -> None:
         """Actually run the comparison.  This tries to run df1.equals(df2)
         first so that if they're truly equal we can tell.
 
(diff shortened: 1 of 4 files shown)


Jin66:
**💬 suggestion :** if this change is accepted, it could be interesting to add a static type check in the CI and some developer documentation about how static typing could be verify locally when updating the library.

With this commit, the following check is ok : 
```
mypy datacompy --strict
```

aguiddir:
**suggestion** 💬  :  If the changes get approved, consider squashing the commits into a single one before merging, as having both might not be necessary

datacompy/core.py
```diff
@@ -275,11 +286,12 @@ def _dataframe_merge(self, ignore_spaces):
         # Clean up temp columns for duplicate row matching
         if self._any_dupes:
             if self.on_index:
-                outer_join.index = outer_join[index_column]
+                # outer_join.index = outer_join[index_column]
```
jdawang, line 289: Can lines 289-291 be further simplified to `outer_join.set_index(index_column, drop=True, inplace=True)`? At least, we should delete the commented-out code.

datacompy/fugue.py
```diff
@@ -406,7 +406,7 @@ def report(
     def shape0(col: str) -> int:
         return sum(x[col][0] for x in res)
 
-    def shape1(col: str) -> int:
+    def shape1(col: str) -> Any:
```
jdawang, line 409: Why is this `Any`?

pyproject.toml
```diff
@@ -94,6 +95,10 @@ dev = [
     "datacompy[build]",
 ]
 
+mypy = [
```
jdawang, line 98: We can just add this to `QA` extras as we will want stubs installed every time mypy is installed.

jdawang (changes requested):
A few comments from me.

datacompy/fugue.py
Jin66: The `res` variable from the the call to `_distributed_compare` is typed as `list[Any]`, then `first` is typed as Any et the shape1 cannot be typed more specifically than Any I believe.

jdawang: Yes, I think you're correct. It's `Any` even though used only in a way that would return `int`.

jdawang (approved):
LGTM, @fdosani?

datacompy/spark.py
```diff
@@ -13,25 +13,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 import sys
 from enum import Enum
 from itertools import chain
+from typing import Any, TextIO, NewType, List, Union, Tuple, Optional, Dict, Set
+
+import pyspark
```
fdosani, line 21: @jdawang @aguiddir @Jin66  I think this breaks the below logic of not mandating pyspark? I might be wrong, but if you don't have pyspark installed, this will still get executed by the `__init__` and throw an `ImportError`

datacompy/spark.py
```diff
@@ -13,25 +13,29 @@
...
 
 try:
     from pyspark.sql import functions as F
 except ImportError:
     pass  # Let non-Spark people at least enjoy the loveliness of the pandas datacompy functionality
 
+SparkDataFrame = pyspark.sql.DataFrame
```
fdosani, line 28: Same as above.

datacompy/fugue.py
aguiddir: In cases where we are certain that the output returns an `int`, we can explicitly cast it to `int` to ensure its type ?

```python
def shape1(col: str) -> int:
    return cast(int, first[col][1])
```

datacompy/spark.py
jdawang: In https://github.com/capitalone/datacompy/blob/3b078e6666a253a96298e1a372fbf419ed6b975a/datacompy/__init__.py#L27

This will raise `ModuleNotFound` error on `import datacompy` if pyspark is a top-level import instead of within the try-except block.

fdosani: If you `import datacompy`, [__init__](https://github.com/capitalone/datacompy/blob/develop/datacompy/__init__.py) is called. In particular:
```python
from datacompy.spark import NUMERIC_SPARK_TYPES, SparkCompare
```
This happens even if you have no intention of using `SparkCompare`. Sorry for not being clearer. Not sure if this make sense, but the `import pyspark` will get executed and if you don't have it installed it will complain in this case. this is why right below we have the `try`, `except`

fdosani: I just tested for my own sanity:

```python
>>> import datacompy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[REDACTED]/datacompy/__init__.py", line 27, in <module>
    from datacompy.spark import NUMERIC_SPARK_TYPES, SparkCompare
  File "[REDACTED]/datacompy/spark.py", line 21, in <module>
    import pyspark
ModuleNotFoundError: No module named 'pyspark'
```

fdosani: @jdawang I like your idea of adding in a test without installing Spark and running the tests

datacompy/fugue.py
aguiddir: ok I leave it as it is currently.

fdosani:
> **suggestion** 💬 : If the changes get approved, consider squashing the commits into a single one before merging, as having both might not be necessary

For sure. Part of the flow is to squash merge into `develop`.

datacompy/spark.py
```diff
@@ -146,9 +146,9 @@ class SparkCompare:
 
     def __init__(
         self,
-        spark_session: SparkSessionType,
-        base_df: SparkDataFrame,
-        compare_df: SparkDataFrame,
+        spark_session: "SparkSessionType",
```
fdosani, line 149: @aguiddir these might need to be "pyspark.sql.SparkSession" and "pyspark.sql.DataFrame" since "SparkSession" and "DataFrame" wouldn't be defined for mypy.

You might also be able to remove 
```python
    SparkSessionType = pyspark.sql.SparkSession
    SparkDataFrame = pyspark.sql.DataFrame
```
from the `try`

datacompy/spark.py
aguiddir: @fdosani, I made the changes to use string version instead of object for typing. Thanks for the suggestion

datacompy/spark.py
aguiddir: @fdosani Done :+1:

datacompy/spark.py
```diff
@@ -531,7 +543,7 @@ def _populate_columns_match_dict(self):
             *self.columns_compared
         )
 
-        def helper(c):
+        def helper(c: str) -> pyspark.sql.Column:
```
aguiddir, line 546:
```suggestion
        def helper(c: str) -> "pyspark.sql.Column":
```

datacompy/spark.py
fdosani: Thank you! Just one more question. Runing mypy gives me this: 
```
datacompy/spark.py:22: error: Cannot find implementation or library stub for module named "pyspark"  [import-not-found]
datacompy/spark.py:23: error: Cannot find implementation or library stub for module named "pyspark.sql.functions"  [import-not-found]
datacompy/spark.py:23: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
datacompy/spark.py:23: error: Cannot find implementation or library stub for module named "pyspark.sql"  [import-not-found]
datacompy/spark.py:463: error: Unused "type: ignore" comment  [unused-ignore]
datacompy/spark.py:472: error: Unused "type: ignore" comment  [unused-ignore]
```

Do we need `type: ignore` on these lines (463 and 472)?

aguiddir: @fdosani Yes, i think we need it because the `orderby` method takes the argument: `*cols: Union[str, Column, List[Union[str, Column]]]`.

The argument we're passing to the method:` self._join_column_names` is of type: `List[str]`.
If we don't address the error, mypy returns this error message:
 `Argument 1 has incompatible type "List[str]"; expected "Union[str, Column, List[Union[str, Column]]]" [arg-type]`

which, in my opinion, isn't a valid error.

I think that to fix this issue, the modification should be made in the 'pyspark' library. The signature of the '[sort](https://github.com/apache/spark/blob/e359318c4493e16a7546d70c9340ffc5015aacff/python/pyspark/sql/dataframe.py#L3036)' method  should change from  `*cols: Union[str, Column, List[Union[str, Column]]]` to ->  `*cols: Union[str, Column, Sequence[Union[str, Column]]]`.

fdosani: Sounds good. Thanks for all you rhelp and patience with these changes etc.

fdosani (approved):
LGTM. Thanks for the diligence!

datacompy/spark.py
aguiddir: Here is the mypy [documentation](https://mypy.readthedocs.io/en/stable/common_issues.html#variance) that explains a bit more about where our problem is coming from.  We can also try to make corrections on our side, but wouldn't it be better to use` # type: ignore[arg-type]` ?

aguiddir:
@fdosani Just one more question : 
How do you merge your PRs? Does it need approval from all 5 reviewers?

fdosani:
> @fdosani Just one more question : 
> 
> How do you merge your PRs? Does it need approval from all 5 reviewers?

No just Jacob and I are fine. Sorry was away from my computer. Will merge in a bit.

PR #251, opened 2023-12-07
New action for running core tests without PySpark

### LINKED ISSUE #250: Add new action for running tests when PySpark is NOT installed
As the title states, we need to ensure that we can use datacompy without installing Spark.

TLDR:
- `import datacompy` should NOT throw any errors when doing the most basic install.

fdosani:
Fixes #250


fdosani:
@jdawang FYI

jdawang:
I did have something similar for running tests with and without spark but moving some stuff around to allow for running the fugue in #252.

Issue #262, opened 2024-01-22
Are there plans to support Python 3.12.1?


RicardoEscobar:
I can't even install it:

```
...
UPDATING build\lib.win-amd64-cpython-312\pandas/_version.py
      set build\lib.win-amd64-cpython-312\pandas/_version.py to '2.0.2'
      running build_ext
      building 'pandas._libs.algos' extension
      error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for pandas
Failed to build pandas
ERROR: Could not build wheels for pandas, which is required to install pyproject.toml-based projects
```

fdosani:
We will support python 3.12. The toml file just needs updating and also there is the question of spark support which currently I'm not sure works with Python 3.12. Just need to work through those questions first, but at a minimum pandas should be able to work. I can take a look at this tomorrow morning and see the LoE involved.

fdosani:
@RicardoEscobar Seems like Spark still doesn't support 3.12 and the same with Ray (Fugue). So we might be stuck without 3.12 support for now.

mbrooksuk:
Hi, just wondering if you have any update on supporting python 3.12.1?

fdosani:
> Hi, just wondering if you have any update on supporting python 3.12.1?

Hey there, We do want to support it. Right now Spark has some compatibility issues. Are you using it for non-Spark use cases?

mbrooksuk:
Hi, yes, purely in a python notebook

fdosani:
Maybe a compromise might be to allow it for Polars and Pandas only until Spark catches up.

mbrooksuk:
Im very new to python, but datacompy is exactly the tool im looking for!  My company only allows python 3.12.1,  how would you allow it for pandas?

fdosani:
> Im very new to python, but datacompy is exactly the tool im looking for! My company only allows python 3.12.1, how would you allow it for pandas?

I'll need to make a change and cut a new release. Will try and come up with something shortly.

fdosani:
Can you provide a output from: `pip freeze` please. Also I just released `0.12.1` if you want to try pip installing that.

RicardoEscobar:
I solved this issue. I'm preparing my answer to this.
[...]

PR #263, opened 2024-01-23
python 3.11 support

fdosani:
This was in ref: #262. I've added in 3.11 support but it seems like Spark and Ray (Fugue) are not compatible with 3.12  as of yet. 😢 

Otherwise I've just made some minor tweaks and added a matrix to the `README`


fdosani:
@jdawang @ak-gupta friendly ping

jdawang (approved):

PR #275, opened 2024-03-05
Pandas on Spark refactor

fdosani:
So a major update here.

- Deprecating SparkCompare into LegacySparkCompare
- SparkCompare will now implement the Pandas on Spark API to better align with Pandas and Polars.
- Bump up to 0.12.0, Support Pandas 2+ and Spark 3.5+ only.

Closes #274 
Closes #13

PR #277, opened 2024-03-12
bug fix when all columns match but no rows match

### LINKED ISSUE #276: `report` throws an exception when all columns match but no rows match
When two Spark dataframes have all matching columns but zero matching rows, the `report` method of `SparkCompare` throws an exception.  Below is the an example piece of code and the result.
___________
Source Code: 
```
import datacompy as dc
from pyspark.sql import SparkSession

spark = SparkSession.builder.getOrCreate()

df1 = spark.createDataFrame(
    [
        (1, "foo", 1),  
        (2, "bar", 1),
    ],
    ["id", "label", "tmp"],  
)

df2 = spark.createDataFrame(
    [
        (3, "foo", 1), 
        (4, "bar", 1),
    ],
    ["id", "label", "tmp"],
)

comparison_report = dc.SparkCompare(spark, base_df=df1, compare_df=df2, join_columns=["id", "label"], cache_intermediates=True)
comparison_report.report()
```

Result: 
```
****** Column Summary ******
Number of columns in common with matching schemas: 3
Number of columns in common with schema differences: 0
Number of columns in base but not compare: 0
Number of columns in compare but not base: 0

****** Row Summary ******
Number of rows in common: 0
Number of rows in base but not compare: 2
Number of rows in compare but not base: 2
Number of duplicate rows found in base: 0
Number of duplicate rows found in compare: 0

****** Row Comparison ******
Number of rows with some columns unequal: 0
Number of rows with all columns equal: 0


TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
----> 1 comparison_report.report()
/python3.8/site-packages/datacompy/spark.py in report(self, file)
    889         self._merge_dataframes()
    890         self._print_num_of_rows_with_column_equality(file)
--> 891         self._print_row_matches_by_column(file)

/lib/python3.8/site-packages/datacompy/spark.py in _print_row_matches_by_column(self, myfile)
    723             if self.columns_match_dict[key][MatchType.MISMATCH.value]
    724         }
--> 725         columns_fully_matching = {
    726             key: self.columns_match_dict[key]
    727             for key in self.columns_match_dict

/lib/python3.8/site-packages/datacompy/spark.py in <dictcomp>(.0)
    726             key: self.columns_match_dict[key]
    727             for key in self.columns_match_dict
--> 728             if sum(self.columns_match_dict[key])
```

fdosani:
Fixes #276 

@SimonBFrank would you mind pulling down this branch and seeing if it fixes your issue. Long story short, there are no results and the subsequent dictionary with values which get displayed out is not able to generate because of `None` values.

More of a hack since we will be deprecating this legacy Spark implementation (#275)  for a much more readable and logically similar to Pandas version. Just catching the `TypeError` for this situation and putting out `{}` for `columns_with_any_diffs` and `columns_fully_matching`

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.11.0"
+__version__ = "0.11.1"
 
 from datacompy.core import *
 from datacompy.fugue import (
--- a/datacompy/spark.py
+++ b/datacompy/spark.py
@@ -722,18 +722,30 @@ class SparkCompare:
             for key in self.columns_match_dict
             if self.columns_match_dict[key][MatchType.MISMATCH.value]
         }
-        columns_fully_matching = {
-            key: self.columns_match_dict[key]
-            for key in self.columns_match_dict
-            if sum(self.columns_match_dict[key])
-            == self.columns_match_dict[key][MatchType.MATCH.value]
-        }
-        columns_with_any_diffs = {
-            key: self.columns_match_dict[key]
-            for key in self.columns_match_dict
-            if sum(self.columns_match_dict[key])
-            != self.columns_match_dict[key][MatchType.MATCH.value]
-        }
+
+        # corner case: when all columns match but no rows match
+        # issue: #276
+        try:
+            columns_fully_matching = {
+                key: self.columns_match_dict[key]
+                for key in self.columns_match_dict
+                if sum(self.columns_match_dict[key])
+                == self.columns_match_dict[key][MatchType.MATCH.value]
+            }
+        except TypeError:
+            columns_fully_matching = {}
+
+        try:
+            columns_with_any_diffs = {
+                key: self.columns_match_dict[key]
+                for key in self.columns_match_dict
+                if sum(self.columns_match_dict[key])
+                != self.columns_match_dict[key][MatchType.MATCH.value]
+            }
+        except TypeError:
+            columns_with_any_diffs = {}
+        #
+
         base_types = {x[0]: x[1] for x in self.base_df.dtypes}
         compare_types = {x[0]: x[1] for x in self.compare_df.dtypes}
 


SimonBFrank:
When I compare the two dataframes with join columns `["id", "label"]` this is the result:

*Dataframe 1:*
| id    | label | tmp |
|-----|------|-----|
| 1      | foo   | 1      |
| 2      | bar   | 1      |

*Dataframe 2:*
| id    | label | tmp |
|-----|------|-----|
| 3      | foo   | 1      |
| 4      | bar   | 1      |

```
****** Column Summary ******
Number of columns in common with matching schemas: 3
Number of columns in common with schema differences: 0
Number of columns in base but not compare: 0
Number of columns in compare but not base: 0

****** Row Summary ******
Number of rows in common: 0
Number of rows in base but not compare: 2
Number of rows in compare but not base: 2
Number of duplicate rows found in base: 0
Number of duplicate rows found in compare: 0

****** Row Comparison ******
Number of rows with some columns unequal: 0
Number of rows with all columns equal: 0

****** Column Comparison ******
Number of columns compared with some values unequal: 0
Number of columns compared with all values equal: 0

****** Columns with Unequal Values ******
Base Column Name  Compare Column Name  Base Dtype     Compare Dtype  # Matches  # Mismatches
----------------  -------------------  -------------  -------------  ---------  ------------
```

I believe `Number of rows with some columns unequal` and `Number of columns compared with some values unequal` should be 2 since only the values in the column `id` are different. Additionally, `Columns with Unequal Values` should have `id`.

fdosani:
> I believe Number of rows with some columns unequal and Number of columns compared with some values unequal should be 2 since only the values in the column id are different. Additionally, Columns with Unequal Values should have id.

I might not be following right, but I think since none of the join columns match (`["id", "label"]`) in this situation it should all be 0. It is joining on both the fields, not just one.

SimonBFrank:
> > I believe Number of rows with some columns unequal and Number of columns compared with some values unequal should be 2 since only the values in the column id are different. Additionally, Columns with Unequal Values should have id.
> 
> I might not be following right, but I think since none of the join columns match (`["id", "label"]`) in this situation it should all be 0. It is joining on both the fields, not just one.

Whoops, I must've been late and I didn't understand it correctly. LGTM

PR #285, opened 2024-03-25
non-join column unicode fix

### LINKED ISSUE #284: SparkCompare [PARSE_SYNTAX_ERROR] if a non-join column name contains unicode symbols
Hi @fdosani , It seems https://github.com/capitalone/datacompy/issues/280 did not fix the case when a column other than join col has an unicode symbol in its name. This is how to reproduce:
```
from pyspark.sql import SparkSession
import datacompy
spark = SparkSession.builder.getOrCreate()

df1 = spark.createDataFrame(
    [
        (1, "foo", "test"),
        (2, "bar", "test"),
    ],
    ["id", "例", "予測対象日"],
)
df2 = spark.createDataFrame(
    [
        (1, "foo", "test"),
        (2, "baz", "test"),
    ],
    ["id", "例", "予測対象日"],
)

comp = datacompy.SparkCompare(spark, df1, df2, join_columns=["例"])
comp.report()
``` 

I took a look into `spark.py`, and it seems the column name is not escaped in `_print_num_of_rows_with_column_equality` method:
```
    def _print_num_of_rows_with_column_equality(self, myfile: TextIO) -> None:
(...)
        where_cond = " AND ".join(
            [
                "A." + name + "=" + str(MatchType.MATCH.value) # <--- here
                for name in self.columns_compared
            ]
        )
(...)
```

fdosani:
Fixes #284 

Adding in escape ` to ensure unicode chars which are not part of the join columns also work.

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.11.2"
+__version__ = "0.11.3"
 
 from datacompy.core import *
 from datacompy.fugue import (
--- a/datacompy/spark.py
+++ b/datacompy/spark.py
@@ -391,7 +391,10 @@ class SparkCompare:
             base_rows.createOrReplaceTempView("baseRows")
             self.base_df.createOrReplaceTempView("baseTable")
             join_condition = " AND ".join(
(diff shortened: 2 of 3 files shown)


jdawang (approved):

fdosani:
@jdawang sorry needed to bump the version too.

PR #291, opened 2024-04-24
Fix suffix on all mismatches reports

enzorooo:
Hi datacompy team!

I noticed that when using `datacompy.Compare(...).all_mismatches()` it defaults to using the `"_df1"` `"_df2"` strings. Which shouldn't be the case when users have predefined this upon declaring the comparison settings. 

I fixed it in this contribution. Let me know if it works for you.

Thanks!

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -4,3 +4,4 @@
 - Mark Zhou
 - Ian Whitestone
 - Faisal Dosani
+- Lorenzo Mercado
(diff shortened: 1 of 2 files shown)


enzorooo:
I see that some tests have failed. Though, I'm not exactly sure why that is given the revision is small. If you can provide some guidance for me to fix it, I'd be glad to do it on my end. 

Thanks!

fdosani:
@enzorooo First off thank you for the PR. Greatly appreciated. I think it is failing cause the test check the formatting of the output which will not match due to the change. Give me a bit here, I'll swing back to this and take a closer look.

fdosani:
Closing in favour of #293

PR #293, opened 2024-04-26
Fix suffix on all mismatches reports

fdosani:
Ref #291
cc: @enzorooo 

Pushing a new branch with some additional changes:

- Ensuring the df names are properly propagated in the compare 
  - also spark and polars implementation
- small fix for one of the polars tests too with varying dtypes.

PR #294, opened 2024-04-26
Fugue count matching rows

fdosani:
Ref #214 
Adding in count_matching_rows to the fugue implementation.


fdosani:
Re-run after #293. There is a Polars fix in there which needs to be merged into dev first.

ak-gupta (approved):

PR #306, opened 2024-05-30
Fugue sql updates

fdosani:


fdosani:
I think we will pause on the fugue SQL implementation within datacompy. We have a snowpark implementation in the works and I'd like to look into native duckdb as well.

PR #309, opened 2024-06-01
adding in benchmark docs

fdosani:

PR #315, opened 2024-06-20
Release v0.13.0

fdosani:
Release v0.13.0
----------------

### New
- #310 
  - new Compare class called `SparkSQLCompare`
  - renamed SparkCompare -> SparkPandasCompare
  - reorganized all the spark logic into a submodule spark
 - #309

### Bug fix
- #308

### General cleanup and house keeping
- #305 

- edgetest bumps
  - #304 
  - #307 
  - #311 
  - #313


satniks:
Hi @fdosani ,  does SparkSQLCompare writes to the DB while processing the data? When I tried it on the databricks serverless compute for job, I get following error:

[NOT_SUPPORTED_WITH_SERVERLESS] PERSIST TABLE is not supported on serverless compute. SQLSTATE: 0A000

DataFrame.persist() is shown in the call stack.

Looks like databricks serverless compute has lots of limitations. 

Anyway SparkSQLCompare works fine for normal databricks compute cluster so this is non-issue for us. We may not use databricks serverless in near future due to such limitations.

fdosani:
> Hi @fdosani , does SparkSQLCompare writes to the DB while processing the data? When I tried it on the databricks serverless compute for job, I get following error:
> 
> [NOT_SUPPORTED_WITH_SERVERLESS] PERSIST TABLE is not supported on serverless compute. SQLSTATE: 0A000
> 
> DataFrame.persist() is shown in the call stack.
> 
> Looks like databricks serverless compute has lots of limitations.
> 
> Anyway SparkSQLCompare works fine for normal databricks compute cluster so this is non-issue for us. We may not use databricks serverless in near future due to such limitations.

So there are a couple of calls to cache which is maybe it? Yeah I'm not sure I want to support serverless as it would restrict most of the code we have it seems.

satniks:
> So there are a couple of calls to cache which is maybe it? Yeah I'm not sure I want to support serverless as it would restrict most of the code we have it seems.

right, serverless will have lots of restrictions and not a common use case. Current support is great. Thanks again.

gladysteh99 (approved):
LGTM!

mattiazenidb:
Hey, I think you should support Databricks Serverless :) The error above is because .cache() is not supported. But it's not supported because it's not needed. The goal of Serverless is to provide best performance out of the box without the user tuning anything.

I tested and it works, I'll create a pull request to support it.

PS: I work for Databricks

fdosani:
> Hey, I think you should support Databricks Serverless :) The error above is because .cache() is not supported. But it's not supported because it's not needed. The goal of Serverless is to provide best performance out of the box without the user tuning anything.
> 
> 
> 
> I tested and it works, I'll create a pull request to support it.
> 
> 
> 
> PS: I work for Databricks

Hey @mattiazenidb I'll happily accept a PR for that. Thanks for helping out here!

Issue #320, opened 2024-07-10
SparkSQLCompare only checks for instance "pyspark.sql.DataFrame", but not for instance "pyspark.sql.connect.dataframe.DataFrame"


achrusciel:
<img width="1092" alt="image" src="https://github.com/capitalone/datacompy/assets/25480933/3bed0f8c-3c60-40bd-9699-b4566557b0b2">
<img width="354" alt="image" src="https://github.com/capitalone/datacompy/assets/25480933/302973c6-7e3d-4bf4-8449-6bd199b0ad98">
<img width="554" alt="image" src="https://github.com/capitalone/datacompy/assets/25480933/a7dbe0ca-356e-4747-a456-e6441431a557">
<img width="477" alt="image" src="https://github.com/capitalone/datacompy/assets/25480933/de0aa96a-6f9e-433a-91e8-62f8c1468b86">

I am using Databricks pyspark:
![image](https://github.com/capitalone/datacompy/assets/25480933/8c4790d9-d70e-43e4-84d8-8cb8c5621082)

When using "from datacompy.spark.legacy import LegacySparkCompare as SparkCompare" the comparison works well.

fdosani:
@achrusciel which version of datacompy are you using?

achrusciel:
This is my docker image:
<img width="480" alt="image" src="https://github.com/capitalone/datacompy/assets/25480933/afbb6e40-0b44-445d-bed8-e357c3528082">

achrusciel:
Sure, this is the result:
<img width="406" alt="image" src="https://github.com/capitalone/datacompy/assets/25480933/6c240174-e15e-485c-9619-3518abaa5914">

fdosani:
[...]

PR #321, opened 2024-07-10
switching version check to spark session

fdosani:
Fix for #320 

Seems like Databricks does something to the `pyspark.version.__version__`. Switching to the actual spark session for the check in this case.

@achrusciel FYI, in case you can try out this specific branch to make sure it works for you.

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.13.1"
+__version__ = "0.13.2"
 
 import platform
 from warnings import warn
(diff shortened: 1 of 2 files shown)


achrusciel:
I am testing the change right now. The code is now executing and I can see in the Spark UI of the cluster that jobs are being executed, and it works.

This is the first time I am switching from the LegacySparkCompare to the new SparkSQLCompare. The same comparison for about 5 million rows completes in ca. 2 minutes when using LegacySparkCompare and with the new SparkSQLCompare it has completed after 8 minutes successfully.

fdosani:
> million rows completes in ca. 2 minutes when using LegacySparkCompare and with the new SparkSQLCompare it has completed after 8 minutes successfull

This is expected. The legacy version dropped duplicates and had some functionality which the new one has and aligns to the Pandas version. The main thing is the dropping duplicates.

jdawang (approved):

PR #326, opened 2024-08-09
Ruff

jdawang:
I have added ruff as our linter and formatter, made corresponding changes to code and added ruff to pre-commit and GH workflows. In this process, I have enabled the numpy rules which prepares datacompy to upgrade to numpy 2.0. Unfortunately, due to compatibility issues with pyspark.pandas and numpy 2.0, some further changes to the imports/guarding will be required to allow numpy 2.0 to be installed alongside datacompy

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -13,6 +13,20 @@ permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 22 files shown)


fdosani (approved):
LGTM

PR #335, opened 2024-09-29
check for is_string_dtype and unsupported mixed type

### LINKED ISSUE #327: pandas df with datetime.date cols returns false differences if ignore_case=True
When ignore_case=True, identical values with type datetime.date will be reported as unequal.
This happens because in core.py/columns_equal() the attempt to call str.upper() on the col for case insensitive-comparison will throw an exception. The try/except returns false for any caught exception.

```
import pandas as pd
import datacompy
import datetime

df1 = pd.DataFrame([{'date_field':datetime.date(2024,5,8)}])
df2=df1.copy()

print("incorrectly reports unequal")                            
results = datacompy.Compare(df1=df1,df2=df2,on_index=True, ignore_case=True)
print(results.report())

results = datacompy.Compare(df1=df1,df2=df2,on_index=True, ignore_case=False)
print("Works correctly")                            
print(results.report())
```

fdosani:
Fixes #327 

This is a breaking change in that we would no longer support mixed column types. As one can imagine, that mixed types can get super complex and cumbersome to manage. 

The proposal here is to **NOT** support mixed types moving forward with `v0.14.0`

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -770,6 +770,11 @@ def columns_equal(
     - Non-numeric values (i.e. where np.isclose can't be used) will just
       trigger True on two nulls or exact matches.
 
+    Notes
+    -----
+    As of version ``0.14.0`` If a column is of a mixed data type the compare will
+    default to returning ``False``.
+
     Parameters
     ----------
     col_1 : Pandas.Series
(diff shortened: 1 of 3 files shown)


fdosani:
@jdawang @ak-gupta @gladysteh99 friendly ping :)

datacompy/core.py
```diff
@@ -792,6 +797,15 @@ def columns_equal(
         values don't match.
     """
     compare: pd.Series[bool]
+
+    # short circuit if comparing mixed type columns. We don't want to support this moving forward.
+    if pd.api.types.infer_dtype(col_1).startswith("mixed") or pd.api.types.infer_dtype(
+        col_1
+    ).startswith("mixed"):
```
gladysteh99, line 804: is this supposed to be `pd.api.types.infer_dtype(col_2).startswith("mixed")` instead for the second clause?

gladysteh99 (commented):
The rest lookd good to me!

gladysteh99 (approved):
LGTM!

PR #338, opened 2024-10-16
unique column name summary display

### LINKED ISSUE #334: Add column names to column summary
Would it be possible to provide not only the _number_ of columns that are found in the original and not in the new but also the _names_ of these columns?

For example, in [this report](https://capitalone.github.io/datacompy/polars_usage.html#reports), it would be good to know that the column that is in original but not in new is named "date_fld".

image

fdosani:
update to the `column_summary.txt` rendering so number of unique column count and names gets displayed.

fixes #334

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -606,8 +606,8 @@ class Compare(BaseCompare):
         report += render(
             "column_summary.txt",
             len(self.intersect_columns()),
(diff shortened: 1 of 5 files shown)


gladysteh99 (approved):
LGTM! 

Minor thoughts: can probably add tests in the future to check the printed output

PR #339, opened 2024-10-17
Add diff for join columns

rhaffar:
Small QoL change - sometimes its not immediately obvious which columns in the join_columns are not in the compared dataframes (especially if upper-casing/lower-casing is involved). This should make it more obvious.


fdosani (approved):
LGTM

PR #343, opened 2024-10-25
 Fix compariosn report when one column is all NAs

yozzo:
With situations when one of the column in the comparison was all NA's then this would break the reporting. For some reason when the matching (boolean) of the actual and expected columns happened when there was a categorical value compared to a NA, the result was NA, rather than False, as it would happen for the other elements in the cols compared.
    
This has now been addressed at the intersect rows level which doesn't seem to break the reporting anymore.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -490,7 +490,7 @@ class Compare(BaseCompare):
             column.
         """
         row_cnt = self.intersect_rows.shape[0]
-        col_match = self.intersect_rows[column + "_match"]
+        col_match = self.intersect_rows[column + "_match"].fillna(False)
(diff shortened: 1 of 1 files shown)


fdosani (commented):
Generally this looks good. I'm wondering though if we should `fillna(False)` permanently on `self.intersect_rows[column + "_match"]` and store it.

yozzo:
> Generally this looks good. I'm wondering though if we should `fillna(False)` permanently on `self.intersect_rows[column + "_match"]` and store it.

tried a different implementation that fixes it at the source of the issue

fdosani (approved):
LGTM. Thanks for the fix here.

PR #349, opened 2024-11-05
Fix edgetest action

jdawang:
Title. Needed to also ignore snowflake tests in edgetest config. Also updated the action. Note: edgetest is still failing on polars.


gladysteh99 (approved):
LGTM!

PR #358, opened 2024-11-27
catch exception, column output change

rhaffar:
- Snowflake throws an error for diff calcs on non-numeric columns in the form of a `ProgrammingError`, but it seems on rare occasion it will throw the exact same error with the same message as a `DatabaseError`. Unsure as to what's happening in the Snowflake backend to cause this, but we should catch both to be safe. SnowparkClientException catch has been removed - it was intended to catch general Snowpark exceptions, but these diff failures come from Snowflake itself failing to calculate the diff, not from Snowpark.
- Adding column name display

Will post passing tests as soon as tests are done running

### DIFF
--- a/datacompy/snowflake.py
+++ b/datacompy/snowflake.py
@@ -32,9 +32,8 @@ from ordered_set import OrderedSet
 
 try:
     import snowflake.snowpark as sp
-    from snowflake.connector.errors import ProgrammingError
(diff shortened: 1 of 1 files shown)


rhaffar:
Screenshot 2024-11-27 at 4 30 25 PM

Tests passing

gladysteh99 (approved):
LGTM!

Issue #359, opened 2024-12-08
Moving Fugue to be an optional dependency.


fdosani:
I'm not sure but my hunch is that most folks are not using Fugue, it might be more niche than mainstream. In order to cut down on dependencies I'm proposing that we move it to a optional dependency, similar to Spark.

@jdawang @ak-gupta @gladysteh99 @rhaffar thoughts?

fdosani:
Completed in #360

PR #360, opened 2024-12-09
fugue optional dependency

fdosani:
Ref: #359 

- Initial PR to move Fugue as being an optional dependency.
- More explicit logging
- General cleanup and skipping of tests when module not found.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -134,7 +134,7 @@ jobs:
     - name: Install datacompy
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .[tests,duckdb,polars,dask,ray]
+        python -m pip install .[tests,fugue]
     - name: Test with pytest
       run: |
         python -m pytest tests/ --ignore=tests/test_snowflake.py
--- a/datacompy/logger.py
+++ b/datacompy/logger.py
@@ -1,0 +1,61 @@
+# SPDX-Copyright: Copyright (c) Capital One Services, LLC
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2024 Capital One Services, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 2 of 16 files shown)


pyproject.toml
```diff
@@ -11,7 +11,7 @@ maintainers = [
   { name="Faisal Dosani", email="faisal.dosani@capitalone.com" }
 ]
 license = {text = "Apache Software License"}
-dependencies = ["pandas<=2.2.3,>=0.25.0", "numpy<=2.1.3,>=1.22.0", "ordered-set<=4.1.0,>=4.0.2", "fugue<=0.9.1,>=0.8.7", "polars<=1.12.0,>=0.20.4"]
+dependencies = ["pandas<=2.2.3,>=0.25.0", "numpy<=2.1.3,>=1.22.0", "ordered-set<=4.1.0,>=4.0.2", "polars[pandas]<=1.12.0,>=0.20.4"]
```
ak-gupta, line 14: [non-blocking] do we need the `pandas` extra here? Based on their `pyproject.toml` it looks like the only addition is `pyarrow`

.github/workflows/test-package.yml
```diff
@@ -134,7 +134,7 @@ jobs:
     - name: Install datacompy
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .[tests,duckdb,polars,dask,ray]
+        python -m pip install .[tests,fugue]
```
ak-gupta, line 137: Should we have an environment for testing `datacompy` without `fugue` installed? Perhaps for an extra step of validation to make sure it's truly optional. Same for Snowflake and Spark I guess

pyproject.toml
fdosani: Yeah there is some Pandas functionality which will fail without the `pandas` extras. I ran into this issue.

.github/workflows/test-package.yml
ak-gupta: Oh yep, sorry didn't see that.

ak-gupta (approved):

datacompy/logger.py
```diff
@@ -0,0 +1,61 @@
...
+        The logging level. Defaults to ``logging.INFO``
+
+    Returns
+    -------
+    logging.Logger
+        Returns a Logger obejct which is set with the passed in paramters.
+        Please see the following for more details:
+        https://docs.python.org/2/library/logging.html
+    """
+    logging.basicConfig(format=FORMAT_TO_USE, stream=sys.stdout, level=log_level)
```
clrusby, line 58: It's not a great idea to change the logging format on the global `logging` package (or to set the `captureWarnings` flag).
These are settings which should be applied globally by the user codebase which is importing `datacompy`.
The net result of this change is that everybody's python logging will be set to log at debug level when they first `import datacompy` - which I'm sure is unexpected!
Suggest simply deleting these two lines.

Issue #342, opened 2024-10-24
Drop support for Python 3.9 by the EoY.


fdosani:
We should start planning to drop support for Python 3.9 by EoY. While we don't need to be militant about it, it would be nice to follow along with https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule

@ak-gupta @jdawang @gladysteh99 thoughts?

PR #363, opened 2025-01-03
Remove Python 3.9 support

gliptak:
#342


gliptak:
there were no changes causing the lint/format failure

fdosani (approved):
LGTM. I think the linting issue might be the change from 3.9 to 3.10 and using X | Y vs Union. Will need to tinker with it a bit more.

PR #366, opened 2025-01-06
Convert to union operator

gliptak:
https://docs.astral.sh/ruff/rules/non-pep604-isinstance/

`Note that this results in slower code. Ignore this rule if the performance of an isinstance or issubclass check is a concern, e.g., in a hot loop.`

#363 #365


fdosani:
I've already captured this in #365

gliptak:
don't see `Would reformat: datacompy/snowflake.py` locally

PR #382, opened 2025-02-18
Update Snowflake Dataframe Naming

### LINKED ISSUE #381: bug: Snowpark/Snowflake comparison of tables named the same, but in different schemas doesn't work
First off, thanks so much for maintaining this super useful package! I just tried the Snowpark/Snowflake functionality and have hit a snag. I'm an Analytics Engineer and I often make changes to a table in my dev environment and compare that to the same table in the prod environment to make sure the changes I've made haven't broken things. I tried to do the following:

```
compare = sp.SnowflakeCompare(
    session,
    f"{db}.prod_schema.my_table",
    f"{db}.dev_schema.my_table",
    join_columns=['id']
)
```

I got an error saying 
`An error occurred during comparison: (1105): The DataFrame does not contain the column named MERGE_MY_TABLE.`

I saw this comment in the [PR](https://github.com/capitalone/datacompy/pull/333) to add SnowflakeCompare and I'm wondering if this is what is causing the issue I'm seeing. Even though the tables are different, they have the same name.
> Running a compare with a table against itself creates some weird issues, as Datacompy starts looking for temporary columns belonging to both tables at a later point in the comparison. There is obviously no use case to run datacompy on a single table, but just thought I’d mention this.

When I change the name of one of tables, all works as expected. 

```
compare = sp.SnowflakeCompare(
    session,
    f"{db}.prod_schema.my_table",
    f"{db}.dev_schema.my_table_dev",
    join_columns=['id']
)
```

Would love to know if there is a workaround for this or if this could be fixed. Thanks again!

rhaffar:
In order to allow for comparisons with similarly named tables across different schemas/dbs, the default naming convention for Snowpark DFs from Snowflake Tables has been updated.

Closes #381 

Snowflake tests passing locally

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -18,7 +18,7 @@ Originally started to be something of a replacement for SAS's PROC COMPARE for P
 Then extended to carry that functionality over to Spark Dataframes.
 """
 
-__version__ = "0.16.2"
+__version__ = "0.16.3"
 
 import platform
 from warnings import warn
--- a/datacompy/snowflake.py
+++ b/datacompy/snowflake.py
@@ -163,7 +163,7 @@ class SnowflakeCompare(BaseCompare):
             if len(table_name) != 3:
                 errmsg = f"{df} is not a valid table name. Be sure to include the target db and schema."
                 raise ValueError(errmsg)
-            self.df1_name = df_name.upper() if df_name else table_name[2]
+            self.df1_name = df_name.upper() if df_name else "__".join(table_name)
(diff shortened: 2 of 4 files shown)


fdosani (approved):
LGTM

rhaffar:
> LGTM, wondering if we want to add some tests too?

Yeah I though of this but wanted to avoiding having to hardcode a separate db/schema, one of the quirks of SF testing. I can make additional environment variables and use those instead.

fdosani (approved):

PR #385, opened 2025-02-26
intersection of rows as the datasets have no mutual key/connection

fdosani:
Changes in this PR are for Pandas and Polars only.

- addition of `only_join_columns` which indicates if all the columns are join columns. This is a specific corner case which prompted this PR.
- tweaks to `_intersect_compare` metrics (`row_cnt`, `match_cnt`)where `only_join_columns` corner case is present.
- `intersect_rows_match` should return `False` if the DataFrame is empty.
- `sample_mismatch` needs to account for the corner case situation rather than look for `column + "_match"`
- `all_mismatch` needs to account for the corner case situation rather than look for `column + "_match"`

@rhaffar I've made a couple of more tweaks after hitting some bugs here and there from the last time you checked. Would appreciate another set of eyes here. 

I'm hoping this covers all the weird corner cases where:
1. all columns are join columns
2. there are no intersecting columns to compare on (aside from the join columns. eg: df1: [a, b, c], df2: [a,b, d], join: [a, b])

### DIFF
--- a/datacompy/polars.py
+++ b/datacompy/polars.py
@@ -536,7 +536,7 @@ class PolarsCompare(BaseCompare):
                 + len(self.df2_unq_rows)
             )
             col_match = self.intersect_rows[column]
-            match_cnt = col_match.sum()
+            match_cnt = col_match.count()
             sample_count = min(sample_count, row_cnt - match_cnt)
             sample = pl.concat(
                 [self.df1_unq_rows[[column]], self.df2_unq_rows[[column]]]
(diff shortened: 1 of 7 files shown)


datacompy/polars.py
```diff
@@ -561,6 +593,15 @@ def all_mismatch(self, ignore_matching_cols: bool = False) -> "pl.DataFrame":
                     LOG.debug(
                         f"Column {orig_col_name} is equal in df1 and df2. It will not be added to the result."
                     )
+        if len(match_list) == 0:
```
rhaffar, line 596: Are we ok returning mismatches based on unique rows when this condition is triggered via all match columns being equal and `ignore_matching_cols` being set to True?

rhaffar (commented):
Looks good Faisal, just one question/comment

datacompy/polars.py
fdosani: I think it is fine. In this situation we are deviating from the normal workflow. This corner cases is more of when we are dealing with the join columns, so by definition it will be a "composite key" of all the join columns. So in this case I think it is fine. Would be interested in your thoughts or insights.

rhaffar (approved):
LGTM

PR #387, opened 2025-03-08
Add analyze_join_columns / get_recommended_join_columns helper functions

mahangu:
We use datacompy for regression testing, and for our use-case it can be useful to try and programmatically work out join columns.

Further, when when prototyping a browser-based UI for datacompy, https://github.com/mahangu/datacompy-web-ui, we worked on a method to [recommend join columns](https://github.com/mahangu/datacompy-web-ui/blob/6ff90033b76ce2293a187e1adbcddbd09cbacbd9/src/datacompy_web_ui/core/comparison.py#L183). 

This PR attempts to bring some of that functionality into datacompy core, via two helper functions:

1. `analyze_join_columns`
2. `get_recommended_join_columns`

### DIFF
--- a/datacompy/helper.py
+++ b/datacompy/helper.py
@@ -1,0 +1,127 @@
+"""
+Helper functions for DataComPy.
+
+This module contains standalone utility functions that can be used
+independently of the main comparison classes.
+"""
+
+import logging
+from typing import Any, Dict, List
+
+import pandas as pd
+
+LOG = logging.getLogger(__name__)
+
+
+def analyze_join_columns(
+    df1: pd.DataFrame, df2: pd.DataFrame, uniqueness_threshold=90, allow_nulls=False
+) -> List[Dict[str, Any]]:
+    """Get statistics for potential join columns to help select appropriate join keys.
+
+    Analyzes columns present in both DataFrames to help users select
+    appropriate join keys. Provides statistics about uniqueness and null
+    values, and recommends columns suitable for joining.
+
+    Parameters
+    ----------
+    df1 : pd.DataFrame
+        First DataFrame to compare
+    df2 : pd.DataFrame
+        Second DataFrame to compare
+    uniqueness_threshold : float, default=90
+        Minimum percentage of uniqueness required in both dataframes for a column
+        to be recommended as a join key
+    allow_nulls : bool, default=False
+        Whether to recommend columns containing null values
+
+    Returns
+    -------
+    List[Dict]
+        List of dictionaries containing column statistics:
+            - Column: Column name
+            - Type: Data type
+            - Unique Values: Count and percentage for both frames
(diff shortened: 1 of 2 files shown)


fdosani:
@mahangu Thanks for the PR, sorry for the delay. Taking a look through some of this still. We were discussing some things internally. Would you be able to articulate the use case for `analyze_join_columns` are you finding you sometimes when comparing it is helpful to have this? More out of curiosity from my side.

cc: @ak-gupta @rhaffar

paddymul:
I think I see a usecase.
Many times I want to compare dataframes that I'm not super familiar with.  I could manually look at each column and figure out the best join columns.  I'm going to do this by my own internalized heruistics.  It would be great to have a function taht expresses my join heuristics so I don't have to scan the columns and do that on my own.

It would be even better if different join columns were emitted,  "what's the second best choice" "third".  Then you could have a UI that cycles through the likely candidates.

mahangu:
> Would you be able to articulate the use case for analyze_join_columns are you finding you sometimes when comparing it is helpful to have this? More out of curiosity from my side.

@fdosani Sure! The main use-case we have for this at the moment internally is in a regression test we have created for our internal data transformation system. Users can generate a scratch table from a changed transformation:

e.g -
original: `marketing.page_views`
scratch: `scratch.marketing__page_views`

They can then run a compare test on both tables using datacompy, to see whether their changes have caused any unforeseen regressions. Currently in this compare script, they need to specify the `join_columns` manually. 

Doing this programmatically for them would decrease friction and make this compare step much smoother in two ways:

1) We could use `analyze_join_columns()` to output a list of possible join columns they can choose from.
2) We could allow them to just use `get_recommended_join_columns()` directly if they want to run a quick/first comparison on the default join columns that datacompy recommends.

Of course, they can also specify custom `join_columns` if they want to. Adding this functionality could just make the initial UX a bit smoother for users. We were initially going to add this feature to our internal script but wanted to explore adding it to datacompy core as we feel it could help other datacompy users as well.

> It would be even better if different join columns were emitted, "what's the second best choice" "third". Then you could have a UI that cycles through the likely candidates.

I have tried to do something like this by adding a `uniqueness_score` that we can order by when providing the join columns to the user - https://github.com/capitalone/datacompy/pull/387/files#diff-0d330e20f29f9779d6ec676715299be6bc6505293f7d15323b134c1ccbe2a2e6R137 - but open to other ideas too!

Thank you both for the reviews! 🙏🏾 Looking forward to chatting further.

fdosani:
@ak-gupta do you have any thoughts on this. I'm OK with adding in the helper functions but I'd like to refine the intent here and have a larger discussion on what we need.

fdosani:
There hasn't been any activity on this for a while. Feel free to reopen if needed, but going to close this for now.

PR #388, opened 2025-03-11
intersection of rows as the datasets have no mutual key/connection - Spark/SF

rhaffar:
Adding compare functionality for Spark and Snowflake Compare when join columns are the only comparable columns.

Added Snowflake tests are passing.


rhaffar:
Pyspark testing framework was only added in 3.5.0, which is why some of the builds are failing. Will update testing to avoid using pyspark testing.

fdosani (approved):
LGTM

PR #395, opened 2025-03-21
Pandas PerformanceWarning fix

### LINKED ISSUE #394: PerformanceWarning for Pandas Comparison with >=100 Columns
This is an issue that has been [brought up before](https://github.com/capitalone/datacompy/issues/188), but the original user who raised it closed that issue.

I've found that once you get to 100+ columns of comparison in a `datacompy.Compare` object, the following warning is repeatedly raised:
```
...lib/python3.12/site-packages/datacompy/core.py:362: PerformanceWarning: DataFrame is highly fragmented.  This is usually the result of calling `frame.insert` many times, which has poor performance.  Consider joining all columns at once using pd.concat(axis=1) instead. To get a de-fragmented frame, use `newframe = frame.copy()`
  self.intersect_rows[col_match] = columns_equal(
```

It logs the warning for each column added, so it can get pretty noisy.

Here's a minimally reproducible example:
```
import pandas as pd
import numpy as np

import datacompy

# Issue occurs when column count is high
# Start seeing PerformanceWarning occur at column_count = 100
row_count = 100
column_count = 110
columns = [str(i) for i in range(column_count)]

# Create two data frames with random integers 0-100 based on row_count and column_count
df1 = pd.DataFrame(
    np.random.randint(0, 100, size=(row_count, column_count)),
    columns=columns,
    index=range(0, row_count),
)
df2 = pd.DataFrame(
    np.random.randint(0, 100, size=(row_count, column_count)),
    columns=columns,
    index=range(0, row_count),
)

# Here we see "PerformanceWarning: DataFrame is highly fragmented." warning
compare_results = datacompy.Compare(df1, df2, on_index=True)
```

There's plenty of resources that will tell you to [just suppress/ignore the warning](https://stackoverflow.com/a/76306267).  This [StackOverflow answer](https://stackoverflow.com/a/77047640) has additional context:

> ...the PerformanceWarning is quite simple: once more than 100 variables are created without specifying dtype or in different fashion than pd.concat then the PerfWarn is always shown

[Here's the pandas code that is the cause with the hard-coded value of 100](https://github.com/pandas-dev/pandas/blob/v2.2.3/pandas/core/internals/managers.py#L1392-L1400). This aligns with my testing that I see this once you get above 100 columns.

Rather than suppressing, if you update [line 362](https://github.com/capitalone/datacompy/blob/develop/datacompy/core.py#L362-L369), to the following, the warnings go away:
```
                self.intersect_rows = pd.concat(
                    [
                        self.intersect_rows,
                        columns_equal(
                            self.intersect_rows[col_1],
                            self.intersect_rows[col_2],
                            self.rel_tol,
                            self.abs_tol,
                            ignore_spaces,
                            ignore_case,
                        ).to_frame(name=col_match)
                    ],
                    axis=1
                )
```

fdosani:
Fixes #394 

To help remove `PerformanceWarning` when there are more than 100 columns being compared. [Link](https://github.com/pandas-dev/pandas/blob/v2.0.1/pandas/core/internals/managers.py#L1426-L1434) to the Pandas code in question.

### DIFF
--- a/datacompy/core.py
+++ b/datacompy/core.py
@@ -359,13 +359,19 @@ class Compare(BaseCompare):
                 col_1 = column + "_" + self.df1_name
                 col_2 = column + "_" + self.df2_name
                 col_match = column + "_match"
-                self.intersect_rows[col_match] = columns_equal(
-                    self.intersect_rows[col_1],
-                    self.intersect_rows[col_2],
-                    self.rel_tol,
-                    self.abs_tol,
-                    ignore_spaces,
-                    ignore_case,
+                self.intersect_rows = pd.concat(
+                    [
+                        self.intersect_rows,
+                        columns_equal(
+                            self.intersect_rows[col_1],
+                            self.intersect_rows[col_2],
+                            self.rel_tol,
+                            self.abs_tol,
+                            ignore_spaces,
+                            ignore_case,
+                        ).to_frame(name=col_match),
+                    ],
+                    axis=1,
                 )
                 match_cnt = self.intersect_rows[col_match].sum()
                 max_diff = calculate_max_diff(


rhaffar (approved):
LGTM

PR #398, opened 2025-03-31
Fix Sample Mismatch Performance

rhaffar:
Forcing a Snowpark dataframe collection to check if the match column exists kills performance on large input table/dataframes, likely for the two following reasons:
- No early collections means no chance for optimization/predicate pushdown in SF
- Collection result is huge.

Instead, we check columns existence directly.


fdosani:
@rhaffar Looks good. Did we want to run an offline intergration test on a snowlfake cluster to ensure it works as expected?

rhaffar:
All tests passing 👍

fdosani (approved):

PR #402, opened 2025-04-15
fix: _dataframe_merge ignore spaces

fdosani:
Fix for Pandas Compare's `_dataframe_merge` and the `ignore_spaces` check. Needed an extra condition for strings types in particular.


rhaffar (approved):
LGTM

PR #404, opened 2025-04-17
feat: Compatibility for Array and Lists

### LINKED ISSUE #58: Inequality when comparing two empty numpy arrays
Comparison of two empty `numpy` arrays currently return `False`, which results in showing diffs where there shouldn't be.

This is due to the way `numpy` compares empty arrays.
Running `bool(np.array([]) == np.array([]))` returns `False` and throws this warning:

```
The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
```

Reproduce this bug with:
```
df1 = pd.DataFrame({"some_col": [np.array([]) for _ in range(10)], "id": [i for i in range(10)]})
df2 = pd.DataFrame({"some_col": [np.array([]) for _ in range(10)], "id": [i for i in range(10)]})

pdcompare = datacompy.Compare(df1, df2, join_columns="id")
print(pdcompare.report())
```
output:
```
DataComPy Comparison
--------------------

DataFrame Summary
-----------------

  DataFrame  Columns  Rows
0       df1        2    10
1       df2        2    10

Column Summary
--------------

Number of columns in common: 2
Number of columns in df1 but not in df2: 0
Number of columns in df2 but not in df1: 0

Row Summary
-----------

Matched on: id
Any duplicates on match values: No
Absolute Tolerance: 0
Relative Tolerance: 0
Number of rows in common: 10
Number of rows in df1 but not in df2: 0
Number of rows in df2 but not in df1: 0

Number of rows with some compared columns unequal: 10
Number of rows with all compared columns equal: 0

Column Comparison
-----------------

Number of columns compared with some values unequal: 1
Number of columns compared with all values equal: 1
Total number of values which compare unequal: 10

Columns with Unequal Values or Types
------------------------------------

     Column df1 dtype df2 dtype  # Unequal  Max Diff  # Null Diff
0  some_col    object    object         10         0            0

Sample Rows with Unequal Values
-------------------------------

   id some_col (df1) some_col (df2)
9   9             []             []
0   0             []             []
3   3             []             []
7   7             []             []
5   5             []             []
1   1             []             []
4   4             []             []
2   2             []             []
8   8             []             []
6   6             []             []
```

fdosani:
This should fix #58 

Adding in `compare_column` compatibility for:
- Pandas
  - `np.ndarray`
  - `list`
- Polars
  - `pl.List`
  - `pl.Array`


gladysteh99 (approved):
LGTM! just curious if there could be performance issue since pandas row-wise operation tends to slow things down?

fdosani:
> LGTM! just curious if there could be performance issue since pandas row-wise operation tends to slow things down?

It is a legit question, but in this case if we want to support arrays there might not be a good alternative. We can make a note that it should be used sparingly?

gladysteh99:
> > LGTM! just curious if there could be performance issue since pandas row-wise operation tends to slow things down?
> 
> It is a legit question, but in this case if we want to support arrays there might not be a good alternative. We can make a note that it should be used sparingly?

Yes I think making a note would be helpful!

gladysteh99 (approved):
LGTM!

PR #410, opened 2025-05-09
Fix bug where ID gen fails if first col is not in index

SBMous:
If the first column in the dataframe does not appear in the index columns, this line currently crashes the comparison.

Switching to selecting the first of the index columns fixes the issue.


fdosani (approved):
Thanks for the fix!

PR #411, opened 2025-05-12
comparing across same datatype, different length

rhaffar:
Snowflake/Snowpark datatypes can vary by length - additionally, string representation of those types bakes in these variances. Updating comparisons to compare directly on types to allow for comparing same types with different lengths/precisions.

The datatype returned by `field.datatype` is actually an instance of a datatype class which bakes in length/precision, which is why we compare on the type of the datatype field. We use the datatype field to call the `simple_string` instance method to get a string representation of the datatype, and the type of the datatype field for actual comparison.

All Snowflake tests passing.


fdosani:
@rhaffar thanks for flagging and fixing. LGTM

fdosani (approved):

PR #424, opened 2025-06-29
feat: custom jinja templating and cleanup

### LINKED ISSUE #423: Custom templates for report creation
Hi, I have a use case where i need more flexibility in the report creation. I would love to use my own templates and pass them to the `report()` method.

Additionally, I feel like there is an weird split between the content of the report coming from the `templates` folder and the `report()` method.
Parts like these might be better located in templates but would require logic in the templates:
```python
if any_mismatch:
    report += "Columns with Unequal Values or Types\n"
    report += "------------------------------------\n"
    report += "\n"
```

I know that i could just create a `report()` method in my own project and that I already have all the flexibility which I am describing here. But if there exists any interest in this project, I am happy to submit a PR.

fdosani:
- ability to provide custom jinja templating in the various reporting methods (pandas, polars, spark, snowflake)
- `LegacySparkCompare` and `SparkPandasCompare` removal from library
- removed unused `datacompy/templates/fav_column_summary.txt`
- updated various docs and added in instructions on how to use templating (jinja)

Closes #423


datacompy/base.py
```diff
@@ -154,15 +156,98 @@ def report(
...
+        template_file = template_path.name
+    else:
+        # Default templates directory
+        template_dir = str(Path(__file__).parent / "templates")
+        template_file = str(template_path)
+
+        # Check if template exists in the default location
+        full_path = Path(template_dir) / template_file
+        # Try without adding .j2 extension if it's already there and file doesn't exist
+        if (
```
rhaffar, line 225: Not sure exactly the intent of this if block - if the path with or without the .j2 extension doesn't exist AND the path ends in j2, then we remove the extension (and it back later)?

I think the other issue here is this code assumes the template are .j2, but .jinja is also a valid jinja file extension. Probably any code that involves us modifying the extension shouldn't be needed.

datacompy/base.py
```diff
@@ -154,15 +156,98 @@ def report(
...
+    # Create Jinja2 environment
+    env = Environment(
+        loader=FileSystemLoader(template_dir),
+        autoescape=select_autoescape(),
+        trim_blocks=True,
+        lstrip_blocks=True,
+    )
+
+    # Ensure template has .j2 extension if needed
+    if (
```
rhaffar, line 241: Same comment regarding modifying file extensions

datacompy/base.py
```diff
@@ -178,16 +263,70 @@ def temp_column_name(*dataframes) -> str:
...
+    html_file = Path(html_file)
+    html_file.parent.mkdir(parents=True, exist_ok=True)
+
+    # Create a simple HTML template with the report in a pre tag
+    html_content = f"""<html><head><title>DataComPy Report</title></head><body><pre>{report}</pre></body></html>"""
+    # Save the HTML file
+    html_file.write_text(html_content, encoding="utf-8")
+
+
+def df_to_str(df: Any, sample_count: int | None = None, on_index: bool = False) -> str:
```
rhaffar, line 295: This will imply some default behaviour which may not be intuitive, and may require us to know to modify it when onboarding new compare objects. This might be better implemented at the concrete compare level - thoughts? (And I imagine long-term a solution that doesn't rely on flowing everything back to Pandas to see what our dataframes look like).

rhaffar (commented):
Some general comments on the base functionality changes, will try to make time to look at the compare-specific implementations later

datacompy/base.py
fdosani: @rhaffar how about this: 

```python

        # If file doesn't exist and has .j2 extension, try without it
        if template_file.endswith(".j2"):
            base_path = full_path.with_suffix("")
            if base_path.exists():
                template_file = template_file[:-3]  # Remove .j2 extension
                full_path = base_path
        # If file doesn't exist and doesn't have .j2 extension, try adding it
        elif not full_path.exists():
            j2_path = full_path.with_suffix(full_path.suffix + ".j2")
            if j2_path.exists():
                template_file = f"{template_file}.j2"
                full_path = j2_path
```

datacompy/base.py
fdosani: I think for now maybe this is fine, but we can refactor this out in a v1 implementation. This is effective a form of dispatching.

datacompy/base.py
fdosani: Updated with some cleaner logic.

fdosani:
friendly ping 😄 @rhaffar @ak-gupta @gladysteh99 @jdawang if someone has time.

datacompy/snowflake.py
```diff
@@ -770,94 +768,101 @@ def all_mismatch(self, ignore_matching_cols: bool = False) -> "sp.DataFrame":
...
         match_sample = []
         any_mismatch = False
+
         for column in self.column_stats:
             if not column["all_match"]:
                 any_mismatch = True
                 match_stats.append(
                     {
-                        "Column": column["column"],
-                        f"{self.df1_name} dtype": column["dtype1"],
```
rhaffar, line 856: May want to keep the df name here for clarity

rhaffar (approved):
Changes look ok at a glance but they're pretty numerous so I can't say I've reviewed them too thoroughly

datacompy/snowflake.py
fdosani: I think it gets captured in the jinja templarte via: `mismatch_stats.df1_name + ' dtype',`
https://github.com/capitalone/datacompy/blob/76cd0a387866dcbd73a7d6dcff91277dc98ea1f6/datacompy/templates/report_template.j2#L46-L47

PR #425, opened 2025-07-04
chore: Update pyproject.toml to include polars as an optional dependency

korvalds:
I found that the polars dependency was a bit excess when only using the spark comparison.
So tried to align the usage with the other backend `[spark]` / `[fugue]` to allow for install `datacompy[polars]` optionally.

Update pyproject.toml to include polars as an optional dependency and update README with how to install.

Hope this aligns with the projects wishes and future work 😄


fdosani:
Hi @korvalds Thanks for the PR. I just want to better understand the issue you were facing a bit better. Are they some constraints what you encountered when installing Polars as part of the base installation?

fdosani:
Sorry should also note, that in the current form this will break the package cause polars will not install by default, but will be imported at runtime and throw an exception.

fdosani:
@korvalds thanks for the context here. I think we probably don't want to make polars an optional dependency. It adds unnecessary complexity in terms of the importing logic and I think in this case would be a breaking change for some users. We made spark as optional since it had a larger footprint. In v1 (coming soon) we might just move to not having optional dependencies at all. But happy to take you input when that time comes. I'm going to close the PR for now. Feel free to comment here if you have any questions or concerns.

PR #433, opened 2025-08-06
Initial v1 refactor

fdosani:
Lots of changes in this PR. Would appreciate a through review here as we are changing a lot.

- switching to "Comparators"
- bumping versions. In particular Spark 4.
- reorganizing the structure of the namespace.

I've created a release branch for v1 specifically to contain all the changes in there before pushing to `dev` and then `main`.

### DIFF
--- a/datacompy/snowflake.py
+++ b/datacompy/snowflake.py
@@ -118,6 +118,8 @@ class SnowflakeCompare(BaseCompare):
     ignore_spaces : bool, optional
         Flag to strip whitespace (including newlines) from string columns (including any join
         columns).
+    ignore_case : bool, optional
+        Flag to ignore the case of string columns
 
     Attributes
     ----------
@@ -138,6 +140,7 @@ class SnowflakeCompare(BaseCompare):
         df1_name: str | None = None,
         df2_name: str | None = None,
         ignore_spaces: bool = False,
+        ignore_case: bool = False,
     ) -> None:
         # Validate tolerance parameters first
         self._abs_tol_dict = validate_tolerance_parameter(
@@ -167,11 +170,12 @@ class SnowflakeCompare(BaseCompare):
         self.abs_tol = abs_tol
         self.rel_tol = rel_tol
         self.ignore_spaces = ignore_spaces
+        self.ignore_case = ignore_case
         self.df1_unq_rows: sp.DataFrame
         self.df2_unq_rows: sp.DataFrame
         self.intersect_rows: sp.DataFrame
         self.column_stats: List[Dict[str, Any]] = []
-        self._compare(ignore_spaces=ignore_spaces)
+        self._compare(ignore_spaces=ignore_spaces, ignore_case=ignore_case)
 
     @property
     def df1(self) -> "sp.DataFrame":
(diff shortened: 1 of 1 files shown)


datacompy/comparator/string.py
```diff
@@ -0,0 +1,594 @@
...
+            col2 = polars_normalize_string_column(
+                col2, self.ignore_space, self.ignore_case
+            )
+            try:
+                return pl.Series(
+                    (col1.eq_missing(col2)) | (col1.is_null() & col2.is_null())
+                )
+            except Exception:
+                try:
+                    return pl.Series(col1.cast(pl.String) == col2.cast(pl.String))
```
gladysteh99, line 143: is there a reason why we use `==` instead of `eq_missing` here?

datacompy/comparator/string.py
```diff
@@ -0,0 +1,594 @@
...
+        date_column = col_1
+
+    try:  # datetime is inferred
+        return pl.Series(
+            (str_column.str.to_datetime(strict=False).eq_missing(date_column))
+            | (str_column.is_null() & date_column.is_null())
+        )
+    except Exception:
+        try:
+            return pl.Series(str_column.cast(pl.String) == date_column.cast(pl.String))
```
gladysteh99, line 592: is there a reason why we use `==` instead of `eq_missing` here?

datacompy/comparator/string.py
```diff
@@ -0,0 +1,594 @@
...
+            The name of the second column to compare.
+        col_match : str
+            The name of the output column that will store the comparison results.
+
+        Returns
+        -------
+        pyspark.sql.DataFrame
+            The DataFrame with an additional column containing the comparison results.
+        None
+            if the columns are not comparable.
```
gladysteh99, line 268: Columns are not comparable if their datatypes are not in any of the ``string/date`` combination.

datacompy/comparator/string.py
```diff
@@ -0,0 +1,594 @@
...
+
+        Raises
+        ------
+        Exception
+            If the comparison fails due to incompatible types or other issues,
+            returns a column of `False` values.
```
gladysteh99, line 274: I think incompatible types would return None..?

datacompy/comparator/string.py
```diff
@@ -0,0 +1,594 @@
...
+        col_match : str
+            The name of the output column that will store the comparison results.
+
+        Returns
+        -------
+        snowflake.snowpark.DataFrame
+            The DataFrame with an additional column containing the comparison results.
+
+        None
+            if the columns are not comparable.
```
gladysteh99, line 347: Same comment as above for ``SparkStringComparator``.

datacompy/comparator/string.py
```diff
@@ -0,0 +1,594 @@
...
+
+        Raises
+        ------
+        Exception
+            If the comparison fails due to incompatible types or other issues,
+            returns a column of `False` values.
```
gladysteh99, line 353: Same comment as above for ``SparkStringComparator``.

datacompy/polars.py
```diff
@@ -1098,32 +1050,32 @@ def generate_id_within_group(
...
-    -----
-    Will not operate on categorical columns.
-    """
-    if str(column.dtype.base_type()) in STRING_TYPE:
-        if ignore_spaces:
-            column = column.str.strip_chars()
-        if ignore_case:
-            column = column.str.to_uppercase()
-    return column
+# def normalize_string_column(
```
gladysteh99, line 1053: remove commented code?

gladysteh99 (commented):
Good work overall! The new separation design looks a lot cleaner. Just a few comments on doc strings and minor details. 

Will take a closer look at the tests on Monday.

datacompy/comparator/string.py
fdosani: Good call, switching this out.

PR #434, opened 2025-08-15
feat: hash sensitive data

TreWSte:
Added two parameters:

- sensitive_data_columns_df1
- sensitive_data_columns_df2

These are lists of column names that indicate what columns contain sensitive data. Those columns then get encrypted, in a way that identical values can still be compared.

### DIFF
--- a/datacompy/utility.py
+++ b/datacompy/utility.py
@@ -1,0 +1,39 @@
+#
+# Copyright 2025 Capital One Services, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Contains utilities for datacompy."""
+
+from functools import wraps
+
+
+def check_module_available(module_available, extra_name):
+    """Create a decorator to check if a module is available.
+
+    This is for runtime checks in the library.
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            if not module_available:
+                raise ImportError(
+                    f"The '{extra_name}' extra is not installed. Please install it to use this functionality, "
+                    f"e.g. `pip install datacompy[{extra_name}]`"
+                )
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
--- a/docs/source/pandas_usage.rst
+++ b/docs/source/pandas_usage.rst
@@ -39,7 +39,7 @@ Set up like:
 
     from io import StringIO
     import pandas as pd
-    import datacompy
+    from datacompy.core import Compare
 
     data1 = """acct_id,dollar_amt,name,float_fld,date_fld
     10000001234,123.45,George Maharis,14530.1555,2017-01-01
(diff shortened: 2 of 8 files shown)


datacompy/core.py
```diff
@@ -156,7 +161,16 @@ def df1(self) -> pd.DataFrame:
 
     @df1.setter
     def df1(self, df1: pd.DataFrame) -> None:
-        """Check that it is a dataframe and has the join columns."""
+        """First, hash any sensitive columns
+        Then, that it is a dataframe and has the join columns."""
+        if self.sensitive_data_columns_df1:
```
fdosani, line 166: update name as per above

datacompy/core.py
```diff
@@ -156,7 +161,16 @@ def df1(self) -> pd.DataFrame:
...
+            for column in self.sensitive_data_columns_df1:
```
fdosani, line 167: update name as per above

datacompy/core.py
```diff
@@ -169,7 +183,16 @@ def df2(self) -> pd.DataFrame:
 
     @df2.setter
     def df2(self, df2: pd.DataFrame) -> None:
-        """Check that it is a dataframe and has the join columns."""
+        """First, hash any sensitive columns
+        Then, that it is a dataframe and has the join columns."""
+        if self.sensitive_data_columns_df2:
```
fdosani, line 188: update name as per above

datacompy/core.py
```diff
@@ -169,7 +183,16 @@ def df2(self) -> pd.DataFrame:
...
+            for column in self.sensitive_data_columns_df2:
```
fdosani, line 189: update name as per above

datacompy/core.py
```diff
@@ -156,7 +161,16 @@ def df1(self) -> pd.DataFrame:
 
     @df1.setter
     def df1(self, df1: pd.DataFrame) -> None:
-        """Check that it is a dataframe and has the join columns."""
+        """First, hash any sensitive columns
+        Then, that it is a dataframe and has the join columns."""
+        if self.sensitive_data_columns_df1:
+            for column in self.sensitive_data_columns_df1:
+                if column in df1.columns:
+                    df1[column] = df1[column].apply(lambda v: hashlib.sha256(str(v).encode('utf-8')).hexdigest())
```
fdosani, line 169: Can we do some analysis on how fast this operation is? I just want to make sure this doesn't turn out to be some bottleneck in terms of performance if there is a lot of data someone ends up hashing. A pure vectorized approach would be the best.

fdosani (changes requested):
Generally good with this. I think we'd want to add this for Polars and Spark too.

- Change `sensitive_data_columns_df1` -> `sensitive_columns_df1`
- Change `sensitive_data_columns_df2` -> `sensitive_columns_df2`
- dig in a bit more on apply here and if it hurts performance vs some other alternative which could be vectorized (if slow)

fdosani:
@TreWSte just wanted to follow up on this PR. No rush.

fdosani:
@TreWSte you might need to rebase your fork to the latest `develop` as there have been some new changes in there since you started working on this. Just an FYI.

datacompy/core.py
```diff
@@ -164,13 +164,9 @@ def df1(self, df1: pd.DataFrame) -> None:
...
-            for column in self.sensitive_columns_df1:
-                if column in df1.columns:
-                    df1[column] = df1[column].apply(lambda v: hashlib.sha256(str(v).encode('utf-8')).hexdigest())
-                else:
-                    raise KeyError(
-                        f"Column name {column} was not found in 'df1'"
-                    )
+            cols_to_hash = [col for col in self.sensitive_columns_df1 if col in df1.columns]
+            if cols_to_hash:
+                df1.loc[:, cols_to_hash] = df1.loc[:, cols_to_hash].astype(str).map(lambda v: hashlib.sha256(v.encode('utf-8')).hexdigest())
```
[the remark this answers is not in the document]
TreWSte: Do you want this updated to your suggestion, or leave as is? I think we should leave it since its faster

Issue #435, opened 2025-08-18
Performance issues with SparkSQLCompare because of nested .withColumn()


korvalds:
While testing SparkSQLCompare with relatively small Spark DataFrames that have many columns (5 rows × 25 columns), I noticed significant performance lags because `.withColumn` is nested inside for-loops (not recommended),

This was my test setup:
```python
from datacompy.spark.sql import SparkSQLCompare
from pyspark.sql import SparkSession
import timeit

spark = SparkSession.builder.getOrCreate()

df1 = spark.createDataFrame([{f"col{i}": f"val{i}_{j}" for i in range(25)} for j in range(5)])
df2 = spark.createDataFrame([
    {f"col{i}": f" val{i}_{j} " if j % (2**i if i < 5 else 1) == 0 else f"val{i}_{j}" for i in range(25)}
    for j in range(5)
])

start = timeit.default_timer()
compare = SparkSQLCompare(spark, df1, df2, "col1", ignore_spaces=False)
compare.intersect_rows.explain(True)
print("Time taken (ignore_spaces=False):", timeit.default_timer() - start)

start = timeit.default_timer()
compare = SparkSQLCompare(spark, df1, df2, "col1", ignore_spaces=True)
compare.intersect_rows.explain(True)
print("Time taken (ignore_spaces=True):", timeit.default_timer() - start)
```

Typical execution plan for this looks like:
```
== Parsed Logical Plan ==
'Project [col0_df1#416, col1_df1#1718, col10_df1#442, col11_df1#468, col12_df1#494, col13_df1#520, col14_df1#546, col15_df1#572, col16_df1#598, col17_df1#624, col18_df1#650, col19_df1#676, col2_df1#702, col20_df1#728, col21_df1#754, col22_df1#780, col23_df1#806, col24_df1#832, col3_df1#858, col4_df1#884, col5_df1#910, col6_df1#936, col7_df1#962, col8_df1#988, ... 53 more fields]
+- Project [col0_df1#416, col1_df1#1718, col10_df1#442, col11_df1#468, col12_df1#494, col13_df1#520, col14_df1#546, col15_df1#572, col16_df1#598, col17_df1#624, col18_df1#650, col19_df1#676, col2_df1#702, col20_df1#728, col21_df1#754, col22_df1#780, col23_df1#806, col24_df1#832, col3_df1#858, col4_df1#884, col5_df1#910, col6_df1#936, col7_df1#962, col8_df1#988, ... 52 more fields]
[...]

PR #438, opened 2025-08-19
Improve performance of SparkSQLCompare by using .withColumns()

fdosani:
Ref PR: #436 
Ref Issue: #435

@korvalds I made a few changes. Appreciate your contributions and if you don't mind reviewing it as well that would be much appreciated.

### changes:
- renamed `columns_equal_expr` to `columns_equal`
- switching string concats with fstrings
- some general house keeping.

I'll try and some some benchmarks too just to make sure the performance overall is still there. @rhaffar wondering if we can also make parallel changes for snowflake in another PR. No rush, just a thought.

### DIFF
--- a/datacompy/spark/sql.py
+++ b/datacompy/spark/sql.py
@@ -753,27 +753,13 @@ class SparkSQLCompare(BaseCompare):
             )
             return to_return
 
-        # expression comparison
-        exprs = {}
-        for c in self.intersect_rows.columns:
-            if c.endswith("_match"):
-                orig_col_name = c[:-6]
-                exprs[c] = columns_equal(
-                    self.intersect_rows,
-                    f"{orig_col_name}_{self.df1_name}",
-                    f"{orig_col_name}_{self.df2_name}",
-                    rel_tol=get_column_tolerance(orig_col_name, self._rel_tol_dict),
-                    abs_tol=get_column_tolerance(orig_col_name, self._abs_tol_dict),
-                    ignore_spaces=self.ignore_spaces,
-                    ignore_case=self.ignore_case,
-                )
+        # get the match cols to aggregate
+        match_cols = [c for c in self.intersect_rows.columns if c.endswith("_match")]
 
-        self.intersect_rows = self.intersect_rows.withColumns(exprs)
-
-        if exprs:
+        if match_cols:
             agg_exprs = [
                 F.sum(F.when(F.col(c) == False, 1).otherwise(0)).alias(f"{c}_count")  # noqa: E712
-                for c in exprs
+                for c in match_cols
             ]
             mismatch_counts = self.intersect_rows.agg(*agg_exprs).first()
         else:


datacompy/spark/sql.py
```diff
@@ -720,24 +748,33 @@ def all_mismatch(
...
+                    self.intersect_rows,
+                    f"{orig_col_name}_{self.df1_name}",
+                    f"{orig_col_name}_{self.df2_name}",
                     rel_tol=get_column_tolerance(orig_col_name, self._rel_tol_dict),
                     abs_tol=get_column_tolerance(orig_col_name, self._abs_tol_dict),
                     ignore_spaces=self.ignore_spaces,
                     ignore_case=self.ignore_case,
                 )
 
+        col_comparison = self.intersect_rows.withColumns(exprs)
```
korvalds, line 771: Is the whole `columns_equal` comparison needed in this method as well? Cause in `_intersect_compare` it is now: `self.intersect_rows = self.intersect_rows.withColumns(exprs)`. Maybe it can just be removed from here.

datacompy/spark/sql.py
```diff
@@ -720,24 +748,33 @@ def all_mismatch(
...
+
+        # metric calculation
+        for c in self.intersect_rows.columns:
+            if c.endswith("_match"):
+                orig_col_name = c[:-6]
+
                 if not ignore_matching_cols or (
                     ignore_matching_cols
                     and col_comparison.select(c)
```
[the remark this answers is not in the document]
fdosani: @korvalds reverted.

datacompy/spark/sql.py
fdosani: i think we can remove it. But we would still need to do the mismatch aggregation correct?
`_intesect_compare()` has:

```python
        if exprs:
            agg_exprs = [
                F.sum(F.when(F.col(c) == True, 1).otherwise(0)).alias(f"{c}_count")  # noqa: E712
                for c in exprs
            ]
            match_counts = self.intersect_rows.agg(*agg_exprs).first()
        else:
            match_counts = {}
```

match_counts isn't saved, so maybe we just calculate mismatches here and save it in `self.column_stats`?
then we can forgo the calculations in `all_mismatch()`

fdosani: @korvalds take a look at this: https://github.com/capitalone/datacompy/pull/438/commits/916cc95f36b2980f77b7c883292a24a87de1cc56 removed the `columns_equal` and simplified a bit.

docs/source/benchmark.rst
```diff
@@ -33,23 +33,23 @@ The Data
 The data (base, and compare) we generated was purely synthetic consisting of 10 columns:
```
korvalds, line 33: Just to mention, the reasons I started investigating this performance issue was because I was doing comparisons on tables with ~700 columns, but only 25 rows. I was puzzled by how slow it was.

Maybe selecting a wider table is more realistic, or including a few different column counts would make the charts more meaningful 🙂

PR #442, opened 2025-09-01
bug: polars calculate_max_diff null handling fix

### LINKED ISSUE #441: `calculate_max_diff` not guaranteed to return `float`, might cause error during rendering of report
I believe the code found [here](https://github.com/capitalone/datacompy/blob/44ef38429365aa962bcdad53269a9727617f4521/datacompy/polars.py#L1046) doesn't really do what it appears to be intended to do:

```python
try:
        return cast(
            float, (col_1.cast(pl.Float64) - col_2.cast(pl.Float64)).abs().max()
        )
    except Exception:
        return 0.0
```

Since `cast` doesn't really do anything to the result of `(col_1.cast(pl.Float64) - col_2.cast(pl.Float64)).abs().max()`, there is no guarantee the return value is indeed a float. One example would be if `col_1` or `col_2` contains only `null`. In this case `max()` returns `None`. 

This will then cause the rendering of the template to throw an error because it can't deal with the `NoneType`. 

Any suggestions on how this should be handled?

fdosani:
Fixes a `TypeError` that occurs when rendering a report if a column contains only null values. This was caused by the `calculate_max_diff` function returning `None` instead of a float, which is not handled by the report template.

The fix includes the following changes:
- null values are now replaced with 0 before calculating the maximum absolute difference. This ensures that the function always returns a float.

Fixes #441

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -18,7 +18,7 @@ Originally started to be something of a replacement for SAS's PROC COMPARE for P
 Then extended to carry that functionality over to Spark Dataframes.
 """
 
-__version__ = "0.18.0"
+__version__ = "0.18.1"
 
 import platform
 from warnings import warn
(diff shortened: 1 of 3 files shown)


Naikless (approved):

gladysteh99 (approved):
LGTM!

Issue #440, opened 2025-09-01
Suppress warnings or redefine them


Naikless:
This is a minor thing, but I am using datacompy in a multiprocess workflow which means that the module is imported a lot. Since I am also using Python 3.12 and have not installed snowflake, spark or fugue, I receive a ton of warnings from my processes.

I can of course suppress them manually on my end, but this might not be the most elegant solution.

fdosani:
Yeah, if you are using multiprocessing it will do that with messages. We have those warnings just for the user knowledge. But if they are annoying can you suppress them by just setting the `logger` to only look at `INFO` when you start up your session(s)?

fdosani:
Closing this issue. Please feel free to reopen if neeed.

PR #443, opened 2025-09-02
chore: format pyproject

fdosani:
- Apply consistent formatting to the `pyproject.toml `moving forward.
- also just tweaking the warning in `__init__` to use logging. (ref: https://github.com/capitalone/datacompy/issues/440)


fdosani:
> LGTM! Is there a reason why we are using pyproject-fmt gng forward?

Figured it would be better to just be consistent and lint it. Also when edgetest runs it will modify it.

PR #452, opened 2025-10-21
enhancement: migrate to trusted publisher and pin commits for publish workflow

MSWechsler:
This PR transitions the package publish workflow from Twine to trusted publisher. In addition, it pins action commit SHAs instead of tags, providing security improvements.

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -10,22 +10,22 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # pin@v2.7.0
       with:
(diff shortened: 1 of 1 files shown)


MSWechsler:
Screenshot 2025-10-21 at 9 38 29 PM

.github/workflows/publish-package.yml
```diff
@@ -10,22 +10,22 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # pin@v2.7.0
```
fdosani: @MSWechsler curious about this. Can we not just pin to a version over a hash? Might make this more readable? Same would go for all the other PRs and instance of this.

fdosani (approved):

fdosani:
@MSWechsler @wrslatz Seems like Mend is not progressing or returning anything back. Any suggestions here?

PR #454, opened 2025-10-27
Fix pandas version constraint to prevent numpy binary incompatibility

KRRT7:
## Problem
When installing datacompy in a fresh environment, the dependency resolver installs `pandas==2.1.1` alongside `numpy>=2.2.6`, which causes a binary incompatibility error:

```
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

This prevents any tests from running and breaks imports.

## Root Cause
The current dependency specification allows `pandas>=0.25,<=2.3.3`, but pandas 2.1.1 has pre-compiled C extensions that are incompatible with numpy 2.2.6. Pandas needs to be at least version 2.2.0 to work with numpy 2.2+.

## Changes
- Updated pandas minimum version constraint from `>=0.25` to `>=2.2.0` in `pyproject.toml`
- Removed upper bound to allow flexibility with newer pandas versions
- This ensures binary compatibility between pandas and numpy 2.x series

## Testing
Verified that after this change:
- `uv sync` installs pandas 2.3.3 (compatible version)
- `pytest --collect-only` successfully collects 269 tests without errors
- No import errors when loading datacompy modules

## Environment Tested
- Python: 3.10.18
- Platform: macOS (ARM64)
- Package manager: uv


fdosani:
@KRRT7 Thanks for flagging this and opening up a PR. I just made a new conda env using the following: 

```bash
conda create -n test python=3.10 ipython uv pip conda openjdk=8
```

and then within the env I installed `datacompy`

```bash
> v pip install datacompy                                                                                                                   
Using Python 3.10.19 environment at: miniconda3/envs/dc_test
Resolved 12 packages in 4.56s
Prepared 1 package in 6.89s
Installed 12 packages in 174ms
 + datacompy==0.18.1
 + jinja2==3.1.6
 + markupsafe==3.0.3
 + numpy==2.2.6
 + ordered-set==4.1.0
 + pandas==2.3.3
 + polars==1.33.1
 + pyarrow==22.0.0
 + python-dateutil==2.9.0.post0
 + pytz==2025.2
 + six==1.17.0
 + tzdata==2025.2
```

It seems to resolve fine for me. I do know the error you mentioned, I've experienced this before myself. We are working on a new 1.0 release here: https://github.com/capitalone/datacompy/blob/cb82bdaaa83587847d76aa0871bc6848fa91d043/pyproject.toml#L36-L42
where we are bumping up releases to the latest. not sure if you have any thoughts on this but maybe end users should be pinning on their end to ensure compatible versions? @ak-gupta @rhaffar any thoughts?

PR #461, opened 2025-11-28
implementation of custom comparators

fdosani:
Additional functionality to implement the comparator changes from the last alpha release. 
- Some core changes include the Base Comparator becoming stateless and things like `rtol` and `atol `moving to the `compare `method.
- documentation updates to showcase usage patterns
- the `columns_equal` updates to reflect the stateless change.

### DIFF
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -47,6 +47,13 @@ from datacompy.comparator.string import pandas_normalize_string_column
 LOG = logging.getLogger(__name__)
 
 
+_PANDAS_DEFAULT_COMPARATORS = [
+    PandasArrayLikeComparator(),
+    PandasNumericComparator(),
+    PandasStringComparator(),
+]
+
+
 class PandasCompare(BaseCompare):
     """Comparison class to be used to compare whether two dataframes as equal.
 
@@ -157,12 +164,7 @@ class PandasCompare(BaseCompare):
 
         Custom comparators are placed first, followed by the default ones.
         """
-        default_comparators = [
-            PandasArrayLikeComparator(),
-            PandasNumericComparator(),
-            PandasStringComparator(),
-        ]
-        return self.custom_comparators + default_comparators
+        return self.custom_comparators + _PANDAS_DEFAULT_COMPARATORS
 
     @property
     def df1(self) -> pd.DataFrame:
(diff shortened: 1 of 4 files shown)


datacompy/pandas.py
```diff
@@ -968,39 +978,45 @@ def columns_equal(
...
+    comparators_ = comparators
+    if not comparators_:
+        # If no comparators are passed, behave as before.
+        comparators_ = [
+            PandasArrayLikeComparator(),
+            PandasNumericComparator(),
+            PandasStringComparator(),
+        ]
+
+    for comparator in comparators_:
```
rhaffar: If I'm understanding correctly, we attempt all comparators sequentially until we get a non-zero compare result? I'd personally prefer a more explicit comparator mapping for the following two reasons:

1. Performance: I could see there being a great deal of overhead in comparing in this manner. For the Snowflake compare for example, you now have to make several compare calls per column - since the computation isn't happening locally, we will have to wait for several back and forths between our local machine and the Snowflake Warehouse before settling on a result.
2. It's not immediately obvious which comparator is being used to compare the columns. If a user implements a comparator that overlaps in type with a default comparator and there is an issue in that comparator, instead of seeing an error the user will get a result from one of the default comparators.

fdosani: 1)
So we will compare with the following: 
- custom first (if they exist) followed by the default ones.
- if at any point a comparator returns None it will move to the next one. If it is anything else it wil either return the t/f or whatever value for that comparator, and end there in terms of comparisons.

2) we could log out which comparator we are at if you think that would be helpful? Just so if someone needs to debug what caused that result to happen?

I feel the mapping solution could be good one. I guess for now I was thinking we could try out a sequence based approach of comparators, but I can see the value of making it an explcit mapping. Maybe a combination of both is actually better? Currently pre v1, it's a sequential list of compares, albeit a short one well defined one.

rhaffar: Sequential can work as well, I think the issue that this is boiling down to is what do we do if users have multiple comparators that consume the same datatypes, which the mapping solution doesn't really solve so much as make it impossible. I do think logging could help, maybe even just logging in what order the comparators will execute for users if we stick to a sequential method.

What would a combination of both look like from your viewpoint? I could see something like a map that consumes the column types of both compare columns, mapping to a list of comparators and trying them in order? It won't make it impossible for us to fall back to a default comparator in a way the user might now want, but it makes it easy for us to know if we are falling back to another comparator, at which point we can log something in a very obvious manner.

rhaffar: Yeah that makes sense - then at that point it's up to the user to make sure that they apply conditions to ensure the columns they're comparing respect the conditions of the comparator.

datacompy/pandas.py
```diff
@@ -968,39 +978,45 @@ def columns_equal(
...
+    compare: pd.Series[bool] | None
+
+    comparators_ = comparators
+    if not comparators_:
+        # If no comparators are passed, behave as before.
+        comparators_ = [
+            PandasArrayLikeComparator(),
+            PandasNumericComparator(),
+            PandasStringComparator(),
+        ]
```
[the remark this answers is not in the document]
rhaffar: Those changes look good to me 👍

PR #466, opened 2025-12-16
edgetest and v1 readme updates

fdosani:
- dependency updates
-  noticed to users about v1 and support/v0.19.x moving forward.


rhaffar (approved):
LGTM

PR #468, opened 2025-12-16
current v1.0.0a2 changes for comparators

fdosani:
- we now have a `support/0.19.x` for legacy 0.19.x fixes. 
- this PR brings in the v1 changes into `develop`
- main changes here include the comparator implementation, docs, and general clean up


rhaffar (approved):
LGTM

PR #477, opened 2026-01-09
benchmark logic

fdosani:
Adding in some basic benchmark logic:
- generate some data (various sizes)
- pytest-benchmark against this data.


rhaffar (approved):
LGTM - nice idea, maybe could be used for an R2 blog post, especially if we introduce significant performance changes

PR #479, opened 2026-01-13
Feature/native arrow support added 

MohitSJha:
* PyArrow.py added which handles data comparisons similar to polars.py
* PyArrow comparators added
* _typing.py added to handle ArrowStreamable (generic arrow-based tables) and ArrowArrayLike (generic arrow-based arrays)

### DIFF
--- a/datacompy/base.py
+++ b/datacompy/base.py
@@ -25,6 +25,7 @@ import logging
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict
+import pyarrow as pa
 
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 from ordered_set import OrderedSet
@@ -184,7 +185,10 @@ class BaseCompare(ABC):
 
     def only_join_columns(self) -> bool:
         """Boolean on if the only columns are the join columns."""
-        return set(self.join_columns) == set(self.df1.columns) == set(self.df2.columns)
+        if isinstance(self.df1, pa.Table) or isinstance(self.df2, pa.Table):
+            return set(self.join_columns) == set(self.df1.schema.names) == set(self.df2.schema.names)
+        else:
+            return set(self.join_columns) == set(self.df1.columns) == set(self.df2.columns)
 
 
 def _resolve_template_path(template_name: str) -> tuple[str, str]:
--- a/datacompy/base.py
+++ b/datacompy/base.py
@@ -336,6 +336,9 @@ def df_to_str(df: Any, sample_count: int | None = None, on_index: bool = False) 
     """
     # Handle pandas DataFrame
     if hasattr(df, "to_string"):
+        # Handle PyArrow Table 
+        if isinstance(df, pa.Table):
+            df = df.to_pandas()
         if sample_count is not None and len(df) > sample_count:
             df = df.head(sample_count)
         if not on_index and hasattr(df, "reset_index"):
--- a/datacompy/comparator/array.py
+++ b/datacompy/comparator/array.py
@@ -20,6 +20,8 @@ import logging
 import numpy as np
 import pandas as pd
 import polars as pl
+import pyarrow as pa
+import pyarrow.compute as pc
 
 from datacompy.comparator.base import BaseComparator
 
@@ -196,5 +198,36 @@ class SnowflakeArrayLikeComparator(BaseComparator):
                 col_match,
                 spf.when(when_clause, spf.lit(True)).otherwise(spf.lit(False)),
             )
+        else:
+            return None
+
+class PyArrowArrayLikeComparator(BaseComparator):
+    """Comparator for array-like columns in PyArrow."""
+
+    def compare(self, col1: pa.Array, col2: pa.Array) -> pa.Array | None:
+        """
+        Compare two array like columns for equality.
+
+        Parameters
+        ----------
+        col1 : pa.Array
+            The first PyArrow Array to compare.
+        col2 : pa.Array
+            The second PyArrow Array to compare.
+
+        Returns
+        -------
+        pa.Array
+            A PyArrow Array of booleans indicating whether the values in `col1` and `col2`
+            are equal.
+        None
+            if the columns are not comparable.
+        """
+        if len(col1) != len(col2):
+            return None
+
+        if pa.types.is_list(col1.type) and pa.types.is_list(col2.type):
+            comparison = pc.equal_with_nulls(col1, col2)
+            return comparison
         else:
             return None
--- a/datacompy/comparator/array.py
+++ b/datacompy/comparator/array.py
@@ -20,9 +20,7 @@ import logging
 import numpy as np
 import pandas as pd
 import polars as pl
-from pre_commit import output
 import pyarrow as pa
-import pyarrow.compute as pc
 
 from datacompy.comparator.base import BaseComparator
 
--- a/datacompy/comparator/array.py
+++ b/datacompy/comparator/array.py
@@ -20,6 +20,7 @@ import logging
 import numpy as np
 import pandas as pd
 import polars as pl
+from pre_commit import output
 import pyarrow as pa
 import pyarrow.compute as pc
 
@@ -28,6 +29,13 @@ from datacompy.comparator.base import BaseComparator
 LOG = logging.getLogger(__name__)
 
 POLARS_ARRAY_TYPE = ["List", "Array"]
+PYARROW_LIST_TYPE_IDS = [
+    pa.lib.Type_LIST,
+    pa.lib.Type_LARGE_LIST, 
+    pa.lib.Type_FIXED_SIZE_LIST,
+    pa.lib.Type_LIST_VIEW, 
+    pa.lib.Type_LARGE_LIST_VIEW
+]
 
 try:
     import pyspark as ps
@@ -226,8 +234,11 @@ class PyArrowArrayLikeComparator(BaseComparator):
         if len(col1) != len(col2):
             return None
 
-        if pa.types.is_list(col1.type) and pa.types.is_list(col2.type):
-            comparison = pc.equal_with_nulls(col1, col2)
-            return comparison
+        if col1.type.id in PYARROW_LIST_TYPE_IDS and col2.type.id in PYARROW_LIST_TYPE_IDS:
+            comparison = []
+            for c1, c2 in zip(col1, col2):
+                comparison.append(c1.equals(c2))
+
+            return pa.array(comparison)
         else:
             return None
--- a/datacompy/comparator/numeric.py
+++ b/datacompy/comparator/numeric.py
@@ -20,8 +20,11 @@ import logging
 import numpy as np
 import pandas as pd
 import polars as pl
+import pyarrow as pa
+import pyarrow.compute as pc
 
 from datacompy.comparator.base import BaseComparator
+from datacompy._typing import ArrowArrayLike
 
 NUMERIC_PANDAS_TYPES = [
     "floating",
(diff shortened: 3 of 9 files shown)


datacompy/base.py
```diff
@@ -184,7 +185,10 @@ def report(
 
     def only_join_columns(self) -> bool:
         """Boolean on if the only columns are the join columns."""
-        return set(self.join_columns) == set(self.df1.columns) == set(self.df2.columns)
+        if isinstance(self.df1, pa.Table) or isinstance(self.df2, pa.Table):
+            return set(self.join_columns) == set(self.df1.schema.names) == set(self.df2.schema.names)
```
fdosani: Just out of curiosity is there any difference between `self.df1..column_names` and `self.df1.schema.names`?

datacompy/comparator/array.py
```diff
@@ -198,3 +200,34 @@ def compare(
...
+        pa.Array
+            A PyArrow Array of booleans indicating whether the values in `col1` and `col2`
+            are equal.
+        None
+            if the columns are not comparable.
+        """
+        if len(col1) != len(col2):
+            return None
+
+        if pa.types.is_list(col1.type) and pa.types.is_list(col2.type):
```
fdosani: PyArrow also has `is_fixed_size_list` and `is_large_list` types that we should considered. Thoughts on those?

datacompy/pyarrow.py
```diff
@@ -0,0 +1,883 @@
...
+            and len(self.df2_unq_rows) == 0
+            and self.intersect_rows_match()
+        )
+
+
+def convert_to_arrow(data_object: ArrowStreamable) -> pa.Table:
+    """Convert a streamable dataframe to pyarrow Table."""
+    LOG.info(f"Validating {data_object} dataframe as PyArrow streamable")
+    if hasattr(data_object, "__arrow_c_stream__"):
+        return pa.Table(data_object)
```
rhaffar: Digging into this a bit (just out of curiousity), it seems like it's not recommended to use the PyArrow Table constructor [directly](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html). Seems the recommendation is to import [from_dataframe](https://arrow.apache.org/docs/python/interchange_protocol.html) - looks to support creating a PyArrow table from any dataframe that supports the dataframe interchange protocol. 

Unclear to me if there's actually any issue with this, or if there's some sort of optimization/performance implication though - purely going off the recommendations in their docs.

datacompy/comparator/array.py
```diff
@@ -198,3 +200,34 @@ def compare(
...
+            comparison = pc.equal_with_nulls(col1, col2)
```
fdosani: Where are you getting `equal_with_nulls`? 
I'm getting `AttributeError: module 'pyarrow.compute' has no attribute 'equal_with_nulls'`

fdosani: I'd also think we should wrap this compare in a `try/except`. Just incase assuming  `equal_with_nulls` is a valid function.

datacompy/comparator/numeric.py
```diff
@@ -392,3 +395,83 @@ def compare(
...
+                close = pc.less_equal(diff.cast(pa.float64()), tolerance)
+                nulls_equal = pc.equal(col1, col2)
+                result = pc.or_(close, nulls_equal)
+                return result
+            except Exception:
+                return pa.array([False] * len(col1), type=pa.bool_())
+        else:
+            return None
+        
+def pyarrow_numeric (col: ArrowArrayLike) -> bool:
```
fdosani: Maybe move this to the utility module?

fdosani (changes requested):
@MohitSJha I haven't fully reviewed the PR, just some initial comments. One thing which is missing is unit tests, if you can please add some testing that would be appreciated.

datacompy/pyarrow.py
```diff
@@ -0,0 +1,883 @@
...
+                [col.lower() for col in dataframe.schema.names]
+            )
+
+        if not set(self.join_columns).issubset(dataframe.schema.names):
+            missing_cols = set(self.join_columns) - set(dataframe.schema.names)
+            raise ValueError(
+                f"{index} is missing join columns: {missing_cols}"
+            )
+        
+        if set(len(dataframe.schema.names)) < len(dataframe.schema.names):
```
rhaffar: I don't think this set is intentional

fdosani: typo? should be `len(set(....` similar to polars:         
https://github.com/capitalone/datacompy/blob/238becd8399ec5a0f07941e8d9ec1190552cacbb/datacompy/polars.py#L197

datacompy/pyarrow.py
```diff
@@ -0,0 +1,883 @@
...
+from copy import deepcopy
+import logging
+from typing import Any, Callable, Dict, Iterable, List, Tuple, cast
+
+from ordered_set import OrderedSet
+import pyarrow as pa
+import pyarrow.compute as pc
+import pyarrow.types as pat
+import numpy as np
+from tomlkit import value
```
fdosani: Doesn't seem to be used?

datacompy/pyarrow.py
```diff
@@ -0,0 +1,883 @@
...
+            # custom comparators can ignore what they don't need.
+            compare = comparator.compare(col_1, col_2, **kwargs)
+
+        if compare is not None:
+            LOG.info(
+                f"Using comparator: {comparator.__class__.__name__} for column ({col_1.name}, {col_2.name}) comparison."
+            )
+            return compare
+
+    compare = pa.Array([False] * len(col_1))
```
fdosani: Should have a `return` here, otherwise it will always return `None` implcitly

MohitSJha:
> @MohitSJha I haven't fully reviewed the PR, just some initial comments. One thing which is missing is unit tests, if you can please add some testing that would be appreciated.

I introduced some typos when I was copying over the files to my personal computer. I will put this PR in draft while I work on fixing those and also while I work on adding some unit tests.

datacompy/comparator/array.py
MohitSJha: Thanks for calling this out. I am thinking of adding something similar to POLARS_ARRAY_TYPE to capture all these different types

datacompy/comparator/array.py
MohitSJha: `pc.equal_with_nulls` was something accidentally added while I was moving stuff around. All those instances should be `pc.equal()`

fdosani:
@MohitSJha any updates on this PR?

MohitSJha:
> @MohitSJha any updates on this PR?

I am working on the updates and adding testing right now. Couldn't get to this earlier because cruise changes ended up taking more time than I anticipated.

MohitSJha:
Added pyarrow test cases and tested against that. PR is now ready for review. Sorry for the delay.

datacompy/comparator/numeric.py
MohitSJha: done.

datacompy/base.py
```diff
@@ -25,6 +25,7 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict
+import pyarrow as pa
```
fdosani, line 28: This makes PyArrow a hard dependency for every user of `datacompy`, even those only using Pandas, Polars, or Spark. If a user doesn't have PyArrow installed, importing `datacompy` will fail entirely

datacompy/base.py
```diff
@@ -184,7 +185,10 @@ def report(
 
     def only_join_columns(self) -> bool:
         """Boolean on if the only columns are the join columns."""
-        return set(self.join_columns) == set(self.df1.columns) == set(self.df2.columns)
+        if isinstance(self.df1, pa.Table) or isinstance(self.df2, pa.Table):
```
fdosani, line 188: Same concern as above. This introduces PyArrow-specific branching into the shared base. PyArrowCompare should override `only_join_columns()` instead. This would also remove the need for the `import pyarrow as pa` in `base.py`

datacompy/pyarrow.py
```diff
@@ -0,0 +1,985 @@
...
+
+    @property
+    def df1(self) -> ArrowStreamable:
+        return self._df1
+
+    @df1.setter
+    def df1(self, df1: ArrowStreamable) -> None:
+        self._df1 = convert_to_arrow(df1) if not isinstance(df1, pa.Table) else df1
+        if self.cast_column_names_lower: 
+            self._df1 = df1.rename_columns(
```
fdosani, line 107: This might be a bug? 
After converting the input via convert_to_arrow(df1), the result is stored in self._df1. 
But then the rename_columns call is on the original df1 parameter, not self._df1.

datacompy/pyarrow.py
```diff
@@ -0,0 +1,985 @@
...
+
+    @property
+    def df2(self) -> ArrowStreamable:
+        return self._df2
+
+    @df2.setter
+    def df2(self, df2: ArrowStreamable) -> None:
+        self._df2 = convert_to_arrow(df2) if not isinstance(df2, pa.Table) else df2
+        if self.cast_column_names_lower: 
+            self._df2 = df2.rename_columns(
```
fdosani, line 120: Same issue as above: https://github.com/capitalone/datacompy/pull/479/changes#r3006422644

datacompy/pyarrow.py
```diff
@@ -0,0 +1,985 @@
...
+    PyArrowArrayLikeComparator,
+    PyArrowNumericComparator,
+    PyArrowStringComparator,
+)
+from datacompy.comparator.base import BaseComparator
+from datacompy.comparator.string import pyarrow_normalize_string_column
+from datacompy.comparator.string import DEFAULT_VALUE
+from datacompy.comparator.array import PYARROW_LIST_TYPE_IDS
+
+LOG = logging.getLogger(__name__)
```
fdosani, line 34: `LOG` is imported from `datacompy.base` on line 16 but then immediately overridden by a local `LOG` on line 34. We should remove `LOG` from the base import.

datacompy/pyarrow.py
```diff
@@ -0,0 +1,985 @@
...
+
+from copy import deepcopy
+import functools
+import logging
+from typing import Any, Dict, List, cast
+
+from ordered_set import OrderedSet
+import pyarrow as pa
+import pyarrow.compute as pc
+import pyarrow.dataset as ds
```
fdosani, line 12: `ds` is only used once (line 354: `ds.field("_merge") == "both"`). Consider replacing with `pc.equal(outer_joined["_merge"], "both")` for consistency with the other filter conditions on lines 342 and 347, which already use `pc.equal`. This would eliminate the `pyarrow.dataset` import entirely.

datacompy/pyarrow.py
```diff
@@ -0,0 +1,985 @@
...
+
+        LOG.debug("Selecting df1 unique rows")
+        self.df1_unq_rows = outer_joined.filter(
+            pc.equal(outer_joined["_merge"], "left_only")
+        ).select(df1_cols)
+
+        LOG.debug("Selecting df2 unique rows")
+        self.df2_unq_rows = outer_joined.filter(
+            pc.equal(outer_joined["_merge"], "right_only")
+        ).select(df2_cols)
```
fdosani, line 348: After the merge, unique row columns have suffixed names (e.g., `col_df1`). In the Pandas and Polars implementations, these are renamed back to the original column names so that `df1_unq_rows.columns` matches `df1.columns`. Here, the suffixed names are preserved. This means `df1_unq_rows` and `df2_unq_rows` will have column names like `value_df1` instead of `value`, which could be confusing in reports and downstream usage.

datacompy/comparator/string.py
```diff
@@ -456,6 +534,19 @@ def polars_normalize_string_column(
...
 
+def pyarrow_normalize_string_column(
+    col: ArrowStreamable,
+    ignore_spaces: bool = False,
+    ignore_case: bool = False
+) -> ArrowArrayLike:
+    """Normalize a string column based on ignore_spaces and ignore_case flags."""
+    if pat.is_string(col.type):
+        if ignore_spaces:
+            col = pc.replace_substring_regex(col, r"\s+", "")
```
fdosani, line 545: Thinking we should be using `pc.utf8_trim(col)` or `pc.utf8_trim_whitespace(col)` instead to match Pandas/Polars behaviour. I think the regex might remove all whitespace characters (including spaces within values), not just leading and trailing whitespace.

datacompy/utility.py
```diff
@@ -16,6 +16,8 @@
 """Contains utilities for datacompy."""
 
 from functools import wraps
+from datacompy._typing import ArrowArrayLike
+import pyarrow as pa
```
fdosani, line 20: Same concern as `base.py`. This makes PyArrow a hard dependency for all users since `utility.py` is imported broadly. Consider making this a lazy import or moving `pyarrow_numeric` to the `pyarrow.py` module.

datacompy/utility.py
```diff
@@ -37,3 +39,26 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+def pyarrow_numeric (col: ArrowArrayLike) -> bool:
```
fdosani, line 43:
```suggestion
def pyarrow_numeric(col: ArrowArrayLike) -> bool:
```

fdosani (changes requested):
Thanks for the PR, and sorry for the delay on this. Let me know if there are any question on the feedback! Also might want to run ruff on all the touched files. It is failing in the pipeline.

fdosani:
Will revisit this at some other time.

Issue #484, opened 2026-01-29
Support for pandas 3.0.0 soon?


tidalvirus:
Pandas was updated to 3.0.0 recently - https://pandas.pydata.org/pandas-docs/stable/whatsnew/v3.0.0.html - any rough indications of when datacompy will be able to support it?

(Currently datacompy supports >= 2.3.3)

Thanks!

fdosani:
@tidalvirus this is something we can take a look at shortly. I'd need to understand if there are any material changes with v3 of Pandas, but if not, then we can probably bring this into our v1 release of datacompy.

fdosani:
For reference: https://pandas.pydata.org/docs/user_guide/migration-3-strings.html
It might be a bit more complicated handling both Pandas 2 and 3 in difference situations. 
I'll try and look into this a bit more. There are some nuances in different situations.

fdosani:
Should be in the latest alpha release: https://github.com/capitalone/datacompy/releases/tag/v1.0.0a5

PR #485, opened 2026-01-30
feat: pandas 3 support

fdosani:
In response to #484 

- changes to the actions matrix to test between pandas 2 and 3
- strings in pandas 3 have a critical bug fixed in them, noted here: https://pandas.pydata.org/docs/user_guide/migration-3-strings.html. 
- casting to a string in pandas 2 would also cast `NaN` as the string `nan`, this is now preserved in pandas 3, so there is new logic to account for this with additional use of `fillna(DEFAULT_VALUE)`
- If we cast to a `str` in pandas 2 we will get the string `nan` which will not be filled with the fillna (which is fine and behaves like before). In pandas 3 it will be a real `NaN` and will get filled by the default value which will allow the comparison to work as expected.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -37,6 +37,9 @@ jobs:
         python-version: ["3.10", "3.11"]
         spark-version: [3.5.7]
         pandas-version: ["2.3.3", "3.0.0"]
+        exclude:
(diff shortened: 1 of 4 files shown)


rhaffar (approved):
LGTM

PR #486, opened 2026-01-30
Pandas: Avoid multiple concat

truls-p:
Thanks for a great library!

I have been using the Pandas functionality to compare some DataFrames that have grown over time. I noticed that with a few hundred columns I was spending more time comparing the DataFrames than I was creating them. 😢 

Some profiling showed that `np.copy` was being called repeatedly due to `pd.concat` inside the loop in: https://github.com/capitalone/datacompy/blob/develop/datacompy/pandas.py#L377-L406

The proposed enhancement in `pandas.py` does the `pd.concat` once outside the loop instead.

To show the behavior I also made some tweaks to the benchmarking code to also produce test files with varying number of columns.
Additionally, I changed the code to use LazyFrames and sink_parquet to generate the test data. It still takes about 4-5 min to generate, but uses significantly less memory.

Here is a pytest benchmark comparing two runs (before the changes in `pandas.py` and after the changes),  showing that the large and wide DataFrame (100,000 x 100) went from 9.1s to 5.4s.

datacompy_comparison

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -9,6 +9,16 @@ FOLDER = "data"
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 DATA_DIR = os.path.join(CURRENT_DIR, FOLDER)
 
+# Define test configurations: (size, num_columns)
+TEST_CONFIGS = [
+    (1000, 9),
+    (1000, 50),
+    (1000, 100),
+    (100000, 9),
+    (100000, 50),
+    (100000, 100),
+]
+
 
 def run_polars(base, compare):  # noqa: D103
     comp = datacompy.polars.PolarsCompare(base, compare, join_columns="id")
@@ -22,20 +32,17 @@ def run_pandas(base, compare):  # noqa: D103
     return None
 
 
-@pytest.mark.parametrize("size", [1000, 100000])
-def test_pandas(benchmark, size):  # noqa: D103
-    base = pd.read_parquet(f"{DATA_DIR}/{size}/base/")
-    compare = pd.read_parquet(f"{DATA_DIR}/{size}/compare/")
+@pytest.mark.parametrize("size,num_cols", TEST_CONFIGS)
+def test_pandas(benchmark, size, num_cols):  # noqa: D103
+    data_path = f"{DATA_DIR}/{size}_{num_cols}cols"
+    base = pd.read_parquet(f"{data_path}/base/")
+    compare = pd.read_parquet(f"{data_path}/compare/")
     benchmark.pedantic(target=run_pandas, args=[base, compare], iterations=1, rounds=5)
 
 
-@pytest.mark.parametrize("size", [1000, 100000])
-def test_polars(benchmark, size):  # noqa: D103
-    base = pl.read_parquet(f"{DATA_DIR}/{size}/base/*.parquet")
-    compare = pl.read_parquet(f"{DATA_DIR}/{size}/compare/*.parquet")
+@pytest.mark.parametrize("size,num_cols", TEST_CONFIGS)
+def test_polars(benchmark, size, num_cols):  # noqa: D103
+    data_path = f"{DATA_DIR}/{size}_{num_cols}cols"
+    base = pl.read_parquet(f"{data_path}/base/*.parquet")
+    compare = pl.read_parquet(f"{data_path}/compare/*.parquet")
     benchmark.pedantic(target=run_polars, args=[base, compare], iterations=1, rounds=5)
-
-
-if __name__ == "__main__":
-    # make sure to install pytest-benchmark
-    retcode = pytest.main(["benchmark.py"])
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -46,3 +46,8 @@ def test_polars(benchmark, size, num_cols):  # noqa: D103
     base = pl.read_parquet(f"{data_path}/base/*.parquet")
     compare = pl.read_parquet(f"{data_path}/compare/*.parquet")
     benchmark.pedantic(target=run_polars, args=[base, compare], iterations=1, rounds=5)
+
+
+if __name__ == "__main__":
+    # make sure to install pytest-benchmark
+    retcode = pytest.main(["benchmark.py"])
--- a/benchmarks/generate_data.py
+++ b/benchmarks/generate_data.py
@@ -1,9 +1,14 @@
-import os  # noqa: D100
+"""Generate benchmark datasets of varying shapes using Polars LazyFrame and sink_parquet."""
+
+import logging
+import os
 import shutil
 import string
 
 import numpy as np
 import polars as pl
+
+logger = logging.getLogger(__name__)
 
 
 def generate_columns(num_columns: int, size: int, seed: int = 42) -> dict:
@@ -141,8 +146,14 @@ def generate_data(
     )
     pl.collect_all(sinks)
 
+
 # typically this is only needs to run once to generate the data
 if __name__ == "__main__":
+
+    logger.basicConfig(
+        level=logging.INFO,
+        format="%(asctime)s - %(levelname)s - %(message)s",
+    )
     folder = "data"
     # delete and recreate the folder
     shutil.rmtree(folder, ignore_errors=True)
@@ -158,5 +169,5 @@ if __name__ == "__main__":
 
     for size in sizes:
         for config in column_configs:
-            print(f"Generating data: size={size}, columns={config['num_base_columns']}")
+            logger.info(f"Generating data: size={size}, columns={config['num_base_columns']}")
             generate_data(size=size, folder=folder, **config)
--- a/benchmarks/generate_data.py
+++ b/benchmarks/generate_data.py
@@ -1,80 +1,162 @@
 import os  # noqa: D100
 import shutil
+import string
 
 import numpy as np
 import polars as pl
-import pyarrow.dataset as ds
 
 
-def generate_data(size=1000, folder="data"):
+def generate_columns(num_columns: int, size: int, seed: int = 42) -> dict:
     """
-    Generate two datasets for benchmarking.
+    Generate a dictionary of column data for benchmarking.
 
-    :param size: dataset size to generate
+    Parameters
+    ----------
+    num_columns : int
+        Number of data columns to generate (excluding id column).
+    size : int
+        Number of rows to generate.
+    seed : int
+        Random seed for reproducibility.
+
+    Returns
+    -------
+    dict
+        Dictionary mapping column names to data arrays.
     """
-    rng = np.random.default_rng()
+    rng = np.random.default_rng(seed)
+    columns = {"id": np.arange(0, size)}
 
-    base = pl.DataFrame(
-        {
-            "id": np.arange(0, size),
-            "a": rng.integers(0, 10, size),
-            "b": rng.random(size),
-            "c": rng.choice(["aaa", "bbb", "ccc"], size),
-            "d": rng.integers(0, 10, size),
-            "e": rng.random(size),
-            "f": rng.choice(["aaa", "bbb", "ccc"], size),
-            "g": rng.integers(0, 10, size),
-            "h": rng.random(size),
-            "i": rng.choice(["aaa", "bbb", "ccc"], size),
-        }
-    ).to_arrow()
+    # Generate column names using alphabet letters, then aa, ab, etc.
+    col_names = list(string.ascii_lowercase)
+    for first in string.ascii_lowercase:
+        for second in string.ascii_lowercase:
+            col_names.append(first + second)
 
-    compare = pl.DataFrame(
-        {
-            "id": np.arange(0, size),
-            "d": rng.integers(0, 10, size),
-            "e": rng.random(size),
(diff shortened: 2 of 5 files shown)


fdosani:
@truls-p appreciate the PR and flagging the issue! Will do a review shortly here. :heart:

benchmarks/generate_data.py
```diff
@@ -1,80 +1,170 @@
...
+        Dictionary mapping column names to data arrays.
+    """
+    rng = np.random.default_rng(seed)
+    columns = {"id": np.arange(0, size)}
+
+    # Generate column names using alphabet letters, then aa, ab, etc.
+    col_names = list(string.ascii_lowercase)
+    for first in string.ascii_lowercase:
+        for second in string.ascii_lowercase:
+            col_names.append(first + second)
```
fdosani:
```suggestion
    col_names + [f"{x}{y}" for x, y in product(col_names, repeat=2)]
```

Wonder if something like the above would be cleaner? Would need to `from itertools import product`

benchmarks/generate_data.py
```diff
@@ -1,80 +1,170 @@
...
+
+    for i in range(num_columns):
+        col_name = col_names[i]
+        col_type = i % 3  # Cycle through int, float, string
+        if col_type == 0:
+            columns[col_name] = rng.integers(0, 10, size)
+        elif col_type == 1:
+            columns[col_name] = rng.random(size)
+        else:
+            columns[col_name] = rng.choice(["aaa", "bbb", "ccc"], size)
```
fdosani:
```suggestion
    generators = [
        lambda: rng.integers(0, 10, size),
        lambda: rng.random(size),
        lambda: rng.choice(["aaa", "bbb", "ccc"], size)
    ]
    columns = {col_names[i]: generators[i % 3]() for i in range(num_columns)}
```

benchmarks/generate_data.py
```diff
@@ -1,80 +1,170 @@
...
-            "k": rng.random(size),
-            "l": rng.choice(["aaa", "bbb", "ccc"], size),
-        }
-    ).to_arrow()
-
-    # write out the datasets as parquet files
-    ds.write_dataset(
-        base,
-        f"{folder}/{size}/base/",
-        basename_template="{i}.parquet",
```
fdosani, line 51: I'd like to preserve this bit of functionality if possible. Having the parquet file broken down into say 20 parts is helpful vs having one single large file. More for distributed Spark since I do some testing offline using a cluster too from time to time.

benchmarks/generate_data.py
```diff
@@ -1,80 +1,170 @@
...
-        max_rows_per_file=1_000_000,
-        max_rows_per_group=1_000_000,
-        existing_data_behavior="overwrite_or_ignore",
+    # Ensure overlap doesn't exceed either dataset's column count
+    overlap_columns = min(overlap_columns, num_base_columns, num_compare_columns)
+
+    # Generate base columns
+    base_cols = generate_columns(num_base_columns, size, seed=42)
+
+    # Generate compare columns with overlapping names for the first overlap_columns
```
fdosani: This seems to be duplicated, any chance we can consolidate this in the function above?

benchmarks/generate_data.py
```diff
@@ -1,80 +1,170 @@
...
+    for first in string.ascii_lowercase:
+        for second in string.ascii_lowercase:
+            col_names.append(first + second)
+
+    # Start column naming from where overlap begins in base
+    base_col_names = list(base_cols.keys())[1:]  # Exclude 'id'
+    overlap_start = num_base_columns - overlap_columns
+    overlap_col_names = base_col_names[overlap_start:]
+
+    # Add overlapping columns (same names, different data)
```
fdosani: duplicate code?

fdosani (commented):
Just some initial thoughts on the PR:
- the changes in `pandas.py` (`concat`) is great! 
- There is a bit of duplication in the benchmark code maybe we can clean up and throw into functions?
- Some minor suggestions for specific code blocks.
Appreciate the help and effort 🚀

truls-p:
> Just some initial thoughts on the PR:
> 
> * the changes in `pandas.py` (`concat`) is great!
> * There is a bit of duplication in the benchmark code maybe we can clean up and throw into functions?
> * Some minor suggestions for specific code blocks.
>   Appreciate the help and effort 🚀

Thanks for all the suggestions! I agree there was some duplicate code in generate_data.py and it was not the most clear... It also was not my intention to remove the chunking of the files, sorry about that!
I have gone through the file and done some refactoring to address your comments and hopefully it is more clear now.

truls-p:
Just a heads up that I removed the `if __name__ == "__main__":` [here ](https://github.com/capitalone/datacompy/pull/486/changes#diff-e4ddaa755e05a39b228e3747b2f3194353baa4c205832cbe314afde37472edb4L37-L41) so that parameters would propagate to pytest-benchmark when running with extra commands, like so `uv run pytest benchmark.py::test_pandas --benchmark-save before`. I noticed that the same `if __name__ == "__main__":` is in [benchmark_spark.py](https://github.com/capitalone/datacompy/pull/486/changes#diff-67455aff71e59b337291446c27e98c683be6218d2e9caf3cd85d8c0e56a1e823L35-L37) as well. It may makes sense to remove it but it depends on what you want to do with these benchmark scripts and how you want to run them.

benchmarks/generate_data.py
```diff
@@ -1,80 +1,179 @@
...
+import os
 import shutil
+import string
+from itertools import product
 
 import numpy as np
 import polars as pl
-import pyarrow.dataset as ds
+
+logger = logging.getLogger(__name__)
```
fdosani, line 13: For some reason I keep getting this error when running the code: `AttributeError: module 'polars' has no attribute 'PartitionBy'. Did you mean: 'PartitionByKey'?`

Figured this might just be easier to implement with a custom `write_partitioned_parquet` function.

```suggestion

def write_partitioned_parquet(
    df: pl.DataFrame,
    output_path: str,
    num_partitions: int = 20,
) -> None:
    """
    Write a DataFrame to multiple parquet files, similar to Spark partitioning.

    Parameters
    ----------
    df : pl.DataFrame
        The DataFrame to write.
    output_path : str
        The output directory path.
    num_partitions : int
        Number of parquet files to create.
    """
    total_rows = df.height
    rows_per_partition = max(1, total_rows // num_partitions)

    for i in range(num_partitions):
        start_idx = i * rows_per_partition
        # Last partition gets remaining rows
        if i == num_partitions - 1:
            end_idx = total_rows
        else:
            end_idx = start_idx + rows_per_partition

        if start_idx >= total_rows:
            break

        partition_df = df.slice(start_idx, end_idx - start_idx)
        partition_file = f"{output_path}/part-{i:05d}.parquet"
        partition_df.write_parquet(partition_file)
 
```

benchmarks/generate_data.py
```diff
@@ -1,80 +1,179 @@
...
+
 
-# typicall this is only needs to run once to generate the data
+def generate_data(
+    size: int = 1000,
+    folder: str = "data",
+    num_base_columns: int = 9,
+    num_compare_columns: int = 9,
+    overlap_columns: int = 6,
+) -> None:
```
fdosani, line 148:
```suggestion
    num_partitions: int = 20,
) -> None:
```

benchmarks/generate_data.py
```diff
@@ -1,80 +1,179 @@
...
+        Dataset size (number of rows) to generate.
+    folder : str
+        Output folder for parquet files.
+    num_base_columns : int
+        Number of data columns for base dataset (excluding id).
+    num_compare_columns : int
+        Number of data columns for compare dataset (excluding id).
+    overlap_columns : int
+        Number of columns that overlap between base and compare datasets.
+    """
```
fdosani, line 166:
```suggestion
    num_partitions : int
        Number of parquet files to create per dataset (similar to Spark partitions).
    """
```

benchmarks/generate_data.py
```diff
@@ -1,80 +1,179 @@
...
-# typicall this is only needs to run once to generate the data
+def generate_data(
+    size: int = 1000,
+    folder: str = "data",
+    num_base_columns: int = 9,
+    num_compare_columns: int = 9,
+    overlap_columns: int = 6,
+) -> None:
+    """
+    Generate two datasets for benchmarking using Polars LazyFrame.
```
fdosani:
```suggestion
    Generate two datasets for benchmarking using Polars.
```

benchmarks/generate_data.py
```diff
@@ -1,80 +1,179 @@
...
+        base_lf.sink_parquet(
+            pl.PartitionBy(f"{base_path}/", max_rows_per_file=100_000), lazy=True
+        )
+    )
+    sinks.append(
+        compare_lf.sink_parquet(
+            pl.PartitionBy(f"{compare_path}/", max_rows_per_file=100_000), lazy=True
+        )
+    )
+    pl.collect_all(sinks)
```
fdosani:
```suggestion
    # Collect and write parquet files with row_group_size for chunking
    base_df = pl.DataFrame(base_columns)
    compare_df = pl.DataFrame(compare_columns)

    # Adjust partitions for small datasets
    effective_partitions = min(num_partitions, size)

    write_partitioned_parquet(base_df, base_path, effective_partitions)
    write_partitioned_parquet(compare_df, compare_path, effective_partitions)
```

benchmarks/benchmark.py
```diff
@@ -22,20 +32,17 @@ def run_pandas(base, compare):  # noqa: D103
...
-    compare = pl.read_parquet(f"{DATA_DIR}/{size}/compare/*.parquet")
+@pytest.mark.parametrize("size,num_cols", TEST_CONFIGS)
+def test_polars(benchmark, size, num_cols):  # noqa: D103
+    data_path = f"{DATA_DIR}/{size}_{num_cols}cols"
+    base = pl.read_parquet(f"{data_path}/base/*.parquet")
+    compare = pl.read_parquet(f"{data_path}/compare/*.parquet")
     benchmark.pedantic(target=run_polars, args=[base, compare], iterations=1, rounds=5)
-
-
-if __name__ == "__main__":
```
fdosani, line 39: can we put this block of code back. this allows a call via: `python benchmark.py` to execute.

fdosani (changes requested):
Couple more suggestions. I was having issues with running the `PartitionBy` for some reason. I'm using Polars: `1.36.1`. thanks again for the effort and changes here.

truls-p:
> Couple more suggestions. I was having issues with running the `PartitionBy` for some reason. I'm using Polars: `1.36.1`. thanks again for the effort and changes here.

`PartitionBy` was introduced in [1.37.0](https://github.com/pola-rs/polars/releases/tag/py-1.37.0) and `PartitionByKey` was deprecated at the same time. Note your [pyproject.toml](https://github.com/capitalone/datacompy/blob/ba3b5d3db10d8b6c955494d8eb8a122ded67afab/pyproject.toml#L41) range for polars includes 1.37.1.

I have added back the `if __name__ == "__main__":`. As for the other suggested changes, I will leave it up to you whether you want to make those changes, but I would expect the memory usage and time it takes to generate the files would increase.

benchmarks/benchmark.py
truls-p: Done, just note that this won't let you pass various [command line arguments to pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options).

benchmarks/generate_data.py
truls-p: As mentioned in my other comment I will leave it up to you whether you want to make this change.

benchmarks/generate_data.py
truls-p: As mentioned in my other comment I will leave it up to you whether you want to make this change.

benchmarks/generate_data.py
truls-p: As mentioned in my other comment I will leave it up to you whether you want to make this change.

benchmarks/generate_data.py
truls-p: As mentioned in my other comment I will leave it up to you whether you want to make this change.

benchmarks/generate_data.py
truls-p: As mentioned in my other comment I will leave it up to you whether you want to make this change.

fdosani:
> > Couple more suggestions. I was having issues with running the `PartitionBy` for some reason. I'm using Polars: `1.36.1`. thanks again for the effort and changes here.
> 
> `PartitionBy` was introduced in [1.37.0](https://github.com/pola-rs/polars/releases/tag/py-1.37.0) and `PartitionByKey` was deprecated at the same time. Note your [pyproject.toml](https://github.com/capitalone/datacompy/blob/ba3b5d3db10d8b6c955494d8eb8a122ded67afab/pyproject.toml#L41) range for polars includes 1.37.1.
> 
> I have added back the `if __name__ == "__main__":`. As for the other suggested changes, I will leave it up to you whether you want to make those changes, but I would expect the memory usage and time it takes to generate the files would increase.

Thanks @truls-p, I think I'd favour the deterministic partitions in this case. Let me make the changes. Thanks for the PR and helping out here to improve the package ❤️

fdosani (approved):

PR #488, opened 2026-02-02
feat: cols with mismatches

### LINKED ISSUE #487: feat: columns with mismatches
Potential feature to return a list of column names where at least one row has a mismatch between df1 and df2. This is useful for identifying problematic columns and potentially rerunning comparisons on a subset of the data.

Something similar to: 

```python
def columns_with_mismatches(self) -> List[str]:
        mismatched_columns = [
            col["column"]
            for col in self.column_stats
            if col["unequal_cnt"] > 0 and col["column"] not in self.join_columns
        ]
        return sorted(mismatched_columns)
```

fdosani:
Fixes #487 

Return a list of column names where at least one row has a mismatch between df1 and df2. This is useful for identifying problematic columns and potentially rerunning comparisons on a subset of the data.

### DIFF
--- a/datacompy/base.py
+++ b/datacompy/base.py
@@ -186,6 +186,31 @@ class BaseCompare(ABC):
         """Boolean on if the only columns are the join columns."""
         return set(self.join_columns) == set(self.df1.columns) == set(self.df2.columns)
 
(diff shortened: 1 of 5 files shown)


rhaffar (approved):
LGTM

Issue #409, opened 2025-05-09
feature request: Mask sensitive columns


stephenpardy:
### Issue
Users of datacompy sometimes have sensitive columns in their data (such as account IDs or other join keys). The comparison report will display these columns as-is leading to potential leakage of this information if not handled correctly. Users currently need to mask the sensitive information either before using datacompy or before sending the report. 

### Solution
Allow users to pass in a list of column names and mask those column values before outputing the comparison report, e.g.:

| ACCOUNT_ID | BALANCE |
| 123 | 100.00 |
| 456 | 200.00 |
| 789 | 50.00 |

Becomes:

| ACCOUNT_ID | BALANCE |
| ***** | 100.00 |
| ***** | 200.00 |
| ***** | 50.00 |

### Alternatives
An alternative to masking is to hash values using a secure hashing algorithm before the performing the comparison. Values that match will be hashed to the same hash value.

fdosani:
@stephenpardy been thinking about this a bit more. Are we just thinking about the join columns here? Or any columns. It feels a bit counter intuitive (at least to me) when you have a mismatch or something you would want to be able to see that. If it is masked it would impede a users investigation?

TreWSte:
@stephenpardy I think the hashing would be the way to go. The data would be masked, but you should still be able to see equivalent values in it. I am looking into adding an optional list of columns to mask, because I think that would be helpful.

@fdosani I think the hashing would resolve your concern about being counter intuitive.

TreWSte:
I am working on this PR for it (https://github.com/capitalone/datacompy/pull/434). I still need to work on it before it gets merged, but please let me know if this idea fulfills the goal or not.

fdosani:
Should be in the latest alpha release: https://github.com/capitalone/datacompy/releases/tag/v1.0.0a5

PR #489, opened 2026-02-10
feat: hash sensitive columns in PandasCompare

sh13m:
# Summary
Continuing the work from [434](https://github.com/capitalone/datacompy/pull/434) which addresses issue [409](https://github.com/capitalone/datacompy/issues/409), merged with a new branch based on the latest develop. 

## Key Changes
- 3 new parameters.
  - `sensitive_data_columns_df1`
  - `sensitive_data_columns_df2`
  - `salt` (an optional salt string, used to increase security during hashing)
- Moved hashing functionality into `_validate_dataframe()`, whereas previously in [434](https://github.com/capitalone/datacompy/pull/434), the hashing was done in the setters before `_validate_dataframe()`. This could have caused issues as the column hashing was done before any checks on the dataframe. 
  - Changed hashing algorithm from `SHA256` to `BLAKE2b`, which is optimized for 64-bit platforms and may be more performant, also natively supports salting whereas it was previously done in `SHA256` with a string concatenation. 
  - Masked each column before applying the hash such that null values remain unhashed. 
- Added validation of `sensitive_data_columns` in `_validate_dataframe()`.
  - Sensitive column names are also cast to lowercase based on `cast_column_names_lower`.
  - Non-unique column names will raise an error.
  - Column names not found in the dataframe will raise an error.
- Added new tests for the additional changes made.

## Considerations
- Whether or not we actually need `salt`, it was included in [434](https://github.com/capitalone/datacompy/pull/434) so I have kept it in. 
- Since `df1` and `df2` are set by reference, hashing any sensitive columns will modify the original dataframes. Not sure if this would be of concern and if we would want some solution to somehow preserve the original columns or leave that responsibility to the user.
 
## Future Enhancements
- Add support for Polars, Spark, and Snowflake
  - Polars support should be pretty easy with the only caveat being that there is a distinction between `NaN` and `null` values whereas there isn't in Pandas, I'm thinking we would simply leave both unhashed in Polars. 
  - Spark and Snowflake may be a bit less straightforward and I would need to look into it a bit more.


datacompy/pandas.py
```diff
@@ -199,7 +212,7 @@ def df2(self, df2: pd.DataFrame) -> None:
     def _validate_dataframe(
         self, index: str, cast_column_names_lower: bool = True
     ) -> None:
-        """Check that it is a dataframe and has the join columns.
+        """Check that it is a dataframe and has the join columns, hash any sensitve columns.
```
```suggestion
        """Check that it is a dataframe and has the join columns, hash any sensitive columns.
```

datacompy/pandas.py
```diff
@@ -149,6 +159,9 @@ def __init__(
             self.on_index = False
 
         self._any_dupes: bool = False
+        self.sensitive_columns_df1 = sensitive_columns_df1
+        self.sensitive_columns_df2 = sensitive_columns_df2
+        self.salt = salt.encode("utf-8")
```
fdosani, line 164: We should have some check here. If the encoding fails it should log out and provide some sensible default to continue. ex: If I were to pass in `None` I think it will throw and exception.

Something like: 
```python
def _encode_salt(self, salt: str | bytes | None) -> bytes | None:
    """Encode and validate salt for blake2b hashing.

    Parameters
    ----------
    salt : str, bytes, or None
        Salt value to encode
    
    Returns
    -------
    bytes or None
        Encoded salt, or None if no salt provided

    Raises
    ------
    TypeError
        If salt is not str, bytes, or None

    ValueError
        If salt exceeds maximum length for blake2b
    """
    if salt is None or salt == "":
        return None
    if isinstance(salt, str):
        encoded = salt.encode("utf-8")
    elif isinstance(salt, bytes):
        encoded = salt
    else:
        raise TypeError(
            f"salt must be str, bytes, or None, got {type(salt).__name__}"
        )
    # BLAKE2b with digest_size=32 allows salt up to 16 bytes
    max_salt_len = 16
    if len(encoded) > max_salt_len:
        raise ValueError(
            f"salt exceeds maximum length of {max_salt_len} bytes "
            f"for blake2b with digest_size=32 (got {len(encoded)} bytes)"
        )
    return encoded
```

this could even live in `base.py` for reuse.

datacompy/pandas.py
```diff
@@ -97,6 +98,12 @@ class PandasCompare(BaseCompare):
         Boolean indicator that controls of column names will be cast into lower case
     custom_comparators : list of ``BaseComparator``, optional
         A list of custom comparator classes to use to compare columns.
+    sensitive_columns_df1: list[str], optional
+        A list of the columns in df1 that should have their values hashed to mask sensitive data.
+    sensitive_columns_df2: list[str], optional
+        A list of the columns in df2 that should have their values hashed to mask sensitive data.
```
rhaffar, line 104: I think this can just be 1 parameter, since if you're hashing a column in one dataframe you'll need to hash the same column in the other.

datacompy/pandas.py
```diff
@@ -212,12 +225,42 @@ def _validate_dataframe(
...
+        if sensitive_columns:
+            for col in sensitive_columns:
+                dataframe[col] = dataframe[col].astype(str)
+                dataframe.loc[dataframe[col].notnull(), col] = dataframe.loc[
+                    dataframe[col].notnull(), col
+                ].map(
+                    lambda v: hashlib.blake2b(
+                        v.encode("utf-8"), digest_size=32, salt=self.salt
+                    ).hexdigest()
+                )
```
fdosani: - Combine the blocks together since we are calling `if sensitive_columns:` twice. 
- There are some difference in converting to strings for pandas 2 and 3. There is a small tweak to manage for that too.

```suggestion
        # Check if sensitive columns are unique and exist in the dataframe, then hash them
        if sensitive_columns:
            # Validation
            if len(set(sensitive_columns)) < len(sensitive_columns):
                raise ValueError(
                    f"sensitive_columns_{index} must have unique column names"
                )

            if not set(sensitive_columns).issubset(set(dataframe.columns)):
                missing_cols = set(sensitive_columns) - set(dataframe.columns)
                raise ValueError(
                    f"sensitive_columns_{index} contains columns not in {index}: {missing_cols}"
                )

            # Hashing (only runs if validation passes)
            for col in sensitive_columns:
                # Preserve nulls before converting to string
                # Note: In pandas 2.x, astype(str) converts None/NaN to strings
                # "None"/"nan". This workaround is for pandas 2.x compatibility.
                # In pandas 3+, astype(str) with the new string dtype would preserve
                # nulls, and this explicit handling could be simplified.
                is_null = dataframe[col].isnull()
                dataframe[col] = dataframe[col].astype(str)
                dataframe.loc[~is_null, col] = dataframe.loc[~is_null, col].map(
                    lambda v: hashlib.blake2b(
                        v.encode("utf-8"), digest_size=32, salt=self.salt
                    ).hexdigest()
                )
                # Restore nulls
                dataframe.loc[is_null, col] = pd.NA
```

datacompy/pandas.py
```diff
@@ -97,6 +98,12 @@ class PandasCompare(BaseCompare):
         Boolean indicator that controls of column names will be cast into lower case
     custom_comparators : list of ``BaseComparator``, optional
         A list of custom comparator classes to use to compare columns.
+    sensitive_columns_df1: list[str], optional
+        A list of the columns in df1 that should have their values hashed to mask sensitive data.
+    sensitive_columns_df2: list[str], optional
+        A list of the columns in df2 that should have their values hashed to mask sensitive data.
+    salt: str, optional
+        An optional salt string, used to increase security during hashing.
```
rhaffar, line 106: Probably don't need to expose this as a parameter. I'm not sure if we actually even need to salt our hashes for the Datacompy use case tbh, but if we want to keep the salt I don't think we need to expose a way for people to provide a salt themselves.

Thoughts @fdosani ?

rhaffar (commented):
Good work Sam 🔥 just a few comments.

Also a separate note, I'd go ahead and squash your commit history now - if by some chance we forget to squash and merge to develop, the commit history is going to get flubbed.

Also, `test_sensitive_columns_null` is failing in a few environments - might be related to the installed version of Pandas in those environments.

fdosani (commented):
@sh13m some initial thoughts and ideas. Happy to discuss further to see what makes sense.

datacompy/pandas.py
fdosani: I can imagine someone want this just incase. What was the thought here? Just unnecessary? I guess its just hashing things in the report at the end of the day.

datacompy/pandas.py
sh13m: Is there ever a use case where users have two slightly different dataframes with sensitive columns only existing in one of them? In either case you're right that needing to type in the same column names twice most of the time would be a bit annoying. If we combine the parameters into one list, probably best to ignore any missing sensitive columns in validation and just filter for the ones in the current dataframe.

datacompy/pandas.py
sh13m: Good catch and yeah we'd definitely need some validation, but I'm wondering the same thing as @rhaffar as to whether we even want this parameter exposed in the first place and if the added security of salting is needed for datacompy.

datacompy/pandas.py
sh13m: Just realized, iirc from the one cryptography course I took in uni, the purpose of salting is to slow down attackers, and the salt is supposed to be random each time. Since in this case we're using the same salt for every hash I'm pretty sure this actually does nothing in terms of security LMAO. Probably just want to remove this as an option.

https://en.wikipedia.org/wiki/Salt_(cryptography)

datacompy/pandas.py
fdosani: I like the single list of names which gets applied across both dfs

PR #490, opened 2026-02-11
feat: hash sensitive columns in PandasCompare

sh13m:
# Summary
Commits in [489](https://github.com/capitalone/datacompy/pull/489) were absolutely cooked and it was easier to just cherrypick my changes into a new branch - reopening a new PR here. This is continuing the work from [434](https://github.com/capitalone/datacompy/pull/434) which addresses issue [409](https://github.com/capitalone/datacompy/issues/409), merged with a new branch based on the latest develop. 

## Key Changes
- Combined `sensitive_columns_df1` and `sensitive_columns_df2` into a single parameter `sensitive_columns`. 
   - Implemented as `@property` and `@sensitive_columns.setter`.
- `salt` parameter removed as it would not add any additional security to the hashing for our use cases.
- Moved hashing functionality into `_hash_sensitive_columns()` which is called after `_validate_dataframe()` and before `_compare()`. 
  - Previously in [434](https://github.com/capitalone/datacompy/pull/434), the hashing was done in the `df` setters before `_validate_dataframe()`. This could have caused issues as the column hashing was done before any checks. 
  - Changed hashing algorithm from `SHA256` to `BLAKE2b`, which is optimized for 64-bit platforms and may be more performant. 
  - Sensitive columns not found in the current dataframe are skipped.
  - Masked each column before applying the hash such that null values remain unhashed. 
- Added validation of `sensitive_columns`.
  - Non-string type column names will raise an error.
  - Sensitive column names are cast to lowercase based on `cast_column_names_lower`.
  - Non-unique column names will raise an error.
- Include warnings for the inplace modification of dataframes containing sensitive columns.
- Added/modified unit tests for the additional changes made.

## Future Enhancements
- Add support for Polars, Spark, and Snowflake
  - Polars support should be pretty easy with the only caveat being that there is a distinction between `NaN` and `null` values whereas there isn't in Pandas, I'm thinking we would simply leave both unhashed in Polars. 
  - Spark and Snowflake may be a bit less straightforward and I would need to look into it a bit more.


sh13m:
With considerations to the preserving the original values to help with user investigation. I wondering if doing some form of symmetric key encryption instead of hashing would be a better idea. 
- We can provide a method like `encrypt()` and `decrypt()` which encrypts/decrypts the sensitive columns before running `_compare()` again. 
- If any sensitive columns exist, run `encrypt()` during instantiation instead of `_compare()`.
- In terms of supplying a private key for the encryption, we can randomly generate one during instantiation or allow the user to supply one through a parameter (I'm more leaning towards only having the randomly generated key though).

This way the original data can still be viewed by the user if they desire, just the datatype would be a string instead of whatever it may have been originally, or we could even store the original datatypes before encryption and convert the columns back after decryption (which may speed up the `_compare()` after).

datacompy/pandas.py
```diff
@@ -212,12 +217,46 @@ def _validate_dataframe(
...
+
+            # Hashing (only runs if validation passes)
+            for col in cols_to_hash:
+                # Preserve nulls before converting to string
+                # Note: In pandas 2.x, astype(str) converts None/NaN to strings
+                # "None"/"nan". This workaround is for pandas 2.x compatibility.
+                # In pandas 3+, astype(str) with the new string dtype would preserve
+                # nulls, and this explicit handling could be simplified.
+                is_null = dataframe[col].isnull()
+                dataframe[col] = dataframe[col].astype(str)
```
fdosani: Thoughts on the mutation here to the original data? At a minimum there should be some documentation or a warning that the original datasets are going to be altered. If someone want to use the dataframe after comparing?

datacompy/pandas.py
```diff
@@ -212,12 +217,46 @@ def _validate_dataframe(
...
+                sensitive_columns = [str(col).lower() for col in sensitive_columns]
         else:
             dataframe.columns = pd.Index([str(col) for col in dataframe.columns])
+
+        # Check if sensitive columns are unique and exist in the dataframe, then hash them
+        if sensitive_columns:
+            # Validation
+            if len(set(sensitive_columns)) < len(sensitive_columns):
+                raise ValueError(
+                    f"sensitive_columns_{index} must have unique column names"
```
fdosani: can we change this output. `sensitive_columns_{index} ` -> `sensitive_columns_df1`, which isn't a thing anymore.

fdosani (commented):
@rhaffar @sh13m one more question about the mutating and editing the original dataframe.

datacompy/pandas.py
sh13m: Some ideas:
- We could store the original unmodified columns somewhere but that could introduce significant memory overhead for larger dataframes.
- ~~I think my [idea](https://github.com/capitalone/datacompy/pull/490#issuecomment-3885697144) for doing an encryption instead of hashing would address this concern since users would be able to restore their original dataframes with a decryption.~~
  - ~~Since the encryption and decryption could be pretty computationally expensive, we could consider not having the encryption be the default behaviour during instantiation but instead controlled by an optional `init_encrypted` parameter. Users would still be able to call `encrypt()` later if they want to hide their data.~~

- I'm also okay with just including a warning in the documentation since this feature isn't that high priority if I'm not mistaken ~~and an encryption type solution may be a bit more complicated in design~~. My understanding is that data hiding really is only needed when users would want to share their report, so they could first do their work without hashing and then run a comparison again on the dataframes with `sensitive_columns` set.

datacompy/pandas.py
sh13m: fixed in: https://github.com/capitalone/datacompy/pull/490/commits/39895c185daa41768d27270432f8da08a173357a

datacompy/pandas.py
sh13m: Thought about it a bit more and I realize encrypting the columns most likely won't work. Duplicate values would have to be encrypted differently unless we reuse the private key and IV (which generally is a very bad idea and would make a lot of security folks very mad 🙃).

I'm now leaning more towards leaving the responsibility of restoring the columns to the user. We can add a warning in the documentation and advise users to manually copy any columns they may still need later before comparing with sensitive columns.

datacompy/pandas.py
```diff
@@ -212,12 +217,53 @@ def _validate_dataframe(
...
             raise TypeError(f"{index} must be a pandas DataFrame")
 
+        sensitive_columns = self.sensitive_columns
+
         if cast_column_names_lower:
             dataframe.columns = pd.Index(
                 [str(col).lower() for col in dataframe.columns]
             )
+            if sensitive_columns:
+                sensitive_columns = [str(col).lower() for col in sensitive_columns]
```
fdosani: I feel like this should just be moved to the `__init__` and lower it at

```python
        self.sensitive_columns = sensitive_columns
```

datacompy/pandas.py
fdosani: I think a warning message **for now** would make the most sense. We can iterate on this feature over time and change things around. I don't love the idea of making copies and blowing up the memory foot print.

Let's log out a warning and let people know this is going to modify in place so be careful. Also a warning in the docstrings makes sense.

sh13m: Sounds good. Just spitballing here, but an idea for keeping the original dataframe unmodified could be to only copy the data needed for generating the report. This way hopefully the memory footprint wouldn't be too large in the typical use case. At that point we could also just do a column replacement instead of hashing which is probably a lot faster.

datacompy/pandas.py
```diff
@@ -212,12 +228,51 @@ def _validate_dataframe(
         if not isinstance(dataframe, pd.DataFrame):
             raise TypeError(f"{index} must be a pandas DataFrame")
 
+        sensitive_columns = self.sensitive_columns
```
fdosani: Is there a reason why we are copying `self.sensitive_columns` and not just referencing it as is? I don't think we are modifying it in this method right?

datacompy/pandas.py
```diff
@@ -148,6 +154,16 @@ def __init__(
             ]
             self.on_index = False
 
+        self.sensitive_columns = sensitive_columns
```
fdosani: I think it might make sense to move this into a `@property` and `@sensitive_columns.setter` to clean up the init a bit here. This could live in `base.py` too since this functionality will need to come to the other dataframe types.

datacompy/pandas.py
sh13m: You're right, this was left over back from when `sensitive_columns` still depended on `index`, will remove.

datacompy/pandas.py
sh13m: I concur, will change this and move to `base.py` once I have time.

datacompy/base.py
sh13m, line 62: Since `senstive_columns` needed `cast_column_names_lower` to be moved to `base.py` as well, I think we're able to just access it directly now instead of passing it as an argument in inherited classes.

datacompy/pandas.py
sh13m: @fdosani implemented the changes in https://github.com/capitalone/datacompy/pull/490/commits/866e442ede8380c3b79ed8cb551ccb9fec255853, let me know what you think.

datacompy/base.py
```diff
@@ -35,6 +35,34 @@
 class BaseCompare(ABC):
     """Base comparison class."""
 
+    def __init__(
+        self,
+        cast_column_names_lower: bool = False,
+        sensitive_columns: List[str] | None = None,
+    ) -> None:
+        self._sensitive_columns: List[str] | None = None
```
fdosani: Do you actually need this line?

datacompy/pandas.py
```diff
@@ -113,8 +118,12 @@ def __init__(
         ignore_case: bool = False,
         cast_column_names_lower: bool = True,
         custom_comparators: List[BaseComparator] | None = None,
+        sensitive_columns: List[str] | None = None,
     ) -> None:
-        self.cast_column_names_lower = cast_column_names_lower
+        super().__init__(
```
fdosani: I'm not sure you need to call `super().__init__` here. thought:

```python
self.cast_column_names_lower = cast_column_names_lower
self.sensitive_columns = sensitive_columns
```

should be enough?

fdosani:
@sh13m I'm going to take a look at this a bit closer tomorrow. Just give me some time. Sorry about the back and forth on this. I might push some commits to your PR if that's cool.

I'm thinking of reverting the changes to base and putting things back in Pandas just for now.

datacompy/pandas.py
```diff
@@ -190,16 +199,74 @@ def df2(self) -> pd.DataFrame:
...
+
+                # Ensure all nulls are consistently represented as pd.NA in the hashed column
+                if is_null.any():
+                    df.loc[is_null, col] = pd.NA
+
     def _validate_dataframe(
         self, index: str, cast_column_names_lower: bool = True
     ) -> None:
-        """Check that it is a dataframe and has the join columns.
+        """Check that it is a dataframe and has the join columns, hash any sensitive columns.
```
```suggestion
        """Check that it is a dataframe and has the join columns.
```

datacompy/pandas.py
```diff
@@ -190,12 +199,70 @@ def df2(self) -> pd.DataFrame:
...
+        # Check duplicates
+        duplicates = {c for c, n in Counter(self.sensitive_columns).items() if n > 1}
+        if duplicates:
+            raise ValueError(f"duplicate columns: {duplicates}")
+
+        # Pre-determine target names once to avoid repeated logic in loops
+        target_cols = [
+            col.lower() if self.cast_column_names_lower else col
+            for col in self.sensitive_columns
+        ]
```
rhaffar: I feel like this should happen in the setter - if you check for dupes before this, then ['acct_id', 'ACCT_ID'] would not be considered a dupe when `cast_column_names_lower` is set.

rhaffar (commented):
Great work Sam 🔥 Just one comment from me

datacompy/pandas.py
sh13m: I'm thinking we could move the validation step to its own method similar to how setting the dfs operate, either way works.

fdosani (approved):
LGTM. @rhaffar will wait for you to review @sh13m latest commit.

rhaffar (approved):
LGTM

PR #492, opened 2026-02-27
feat: hash sensitive columns in PolarsCompare

sh13m:
Add sensitive column hashing to PolarsCompare, similar functionality to the existing one in PandasCompare.

- New `sensitive_columns` parameter with same getter and setter implementation + column names validation.
- Polars dataframes are immutable, hashing columns will require an internal copy of the dataframes, warnings updated to reflect this change in behaviour.
- `Nulls` are skipped when hashing values, however polars distinguishes between `Null` and `NaN` values unlike in pandas.  As such, `NaN` values are cast to string and hashed in PolarsCompare.
- Same unit tests + 1 extra for `NaN`.

### DIFF
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -1037,15 +1037,6 @@ def columns_equal(
     - Non-numeric values (i.e. where np.isclose can't be used) will just trigger
       ``True`` on two nulls or exact matches.
 
-    Notes
-    -----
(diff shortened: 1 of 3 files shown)


sh13m:
Going to try to iterate on PandasCompare to remove inplace modification first, may be able to reuse ideas to reduce the internal df copying needed here.

PR #493, opened 2026-03-03
feat: hash sensitive columns in PandasCompare without modifying original dfs

sh13m:
Improves on PR [490](https://github.com/capitalone/datacompy/pull/490), hashing can now be done without affecting original dataframes and should not blow up memory footprint. 

- Realized data is already being copied internally when merging dataframes in `_dataframe_merge()`, so we can hash the relevant columns in `df1_unq_rows`, `df2_unq_rows`, and `intersect_rows` after calling `_compare()` to achieve sensitive column hashing without modifying original dataframes.
- Something not caught previously but should be fixed now is that tolerances wouldn't have worked for floats because the hashing was done prior to comparing.
- Also shouldn't introduce any significant memory pressure since we are modifying the internal dataframes inplace.
- Updated unit tests.

### DIFF
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -101,8 +101,8 @@ class PandasCompare(BaseCompare):
         A list of custom comparator classes to use to compare columns.
     sensitive_columns: list[str], optional
         A list of the columns in df1 or df2 that should have their values hashed to mask sensitive data.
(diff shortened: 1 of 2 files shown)


sh13m:
closed and reopened because new commit wasn't getting added to the pr for some reason, works now

sh13m:
Okay new idea just came to mind today, _should_ be the final iteration for this feature.

PR #494, opened 2026-03-06
feat: hide sensitive columns in PandasCompare without modifying original dfs

sh13m:
Improves on https://github.com/capitalone/datacompy/pull/490. Column hashing changed to data hiding/censoring. Can now be done without affecting the original dataframes and should not blow up memory footprint. 

- Realized data is already being copied internally when merging dataframes in `_dataframe_merge()`, with `df1_unq_rows`, `df2_unq_rows`, and `intersect_rows` being used downstream.
- We can modify these dataframes after `_compare()` is called to hide the relevant sensitive columns without modifying the original dataframes, also shouldn't introduce any significant memory pressure.
- Something not caught previously but should be fixed now is that tolerances wouldn't have worked for floats because the hashing was done prior to comparing.
- Since things are being done after the compare now instead of before, we can change the hashing to a simple column replacement of all `*******` to achieve full data hiding.
    - Much faster and simpler this way (don't need to worry about nulls).
- Sensitive column hiding is now done through the `hide_sensitive_columns(sensitive_columns)` method, we can also reveal the hidden columns through  `reveal_sensitive_columns()`.
    - combined the setter and validation into a single private method to prevent direct attribute modification, `self.sensitive_columns` should only be controlled through the hide/reveal methods.
- Updated unit tests.

### DIFF
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -223,7 +223,7 @@ class PandasCompare(BaseCompare):
         if duplicates:
             raise ValueError(f"duplicate columns: {duplicates}")
 
-        # Warn if column not in both df1 and df2
+        # Warn if column not found in either dataframe
         unused = [
             col
             for col in self.sensitive_columns
@@ -231,7 +231,7 @@ class PandasCompare(BaseCompare):
         ]
         if unused:
             LOG.warning(
-                f"sensitive columns not found in both df1 and df2 will be ignored: {unused}"
+                f"sensitive columns not found in either df1 or df2 will be ignored: {unused}"
             )
 
     def hide_sensitive_columns(self, sensitive_columns: List[str]) -> None:
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -117,7 +117,7 @@ class PandasCompare(BaseCompare):
     ) -> None:
         self.cast_column_names_lower = cast_column_names_lower
         self.custom_comparators = custom_comparators or []
-        self.sensitive_columns: List[str] | None = None
+        self._set_sensitive_columns(None)
 
         # Validate tolerance parameters first
         self._abs_tol_dict = validate_tolerance_parameter(
@@ -203,8 +203,7 @@ class PandasCompare(BaseCompare):
         """Get the list of sensitive columns."""
         return self._sensitive_columns
 
-    @sensitive_columns.setter
-    def sensitive_columns(self, sensitive_columns: List[str] | None) -> None:
+    def _set_sensitive_columns(self, sensitive_columns: List[str] | None) -> None:
         """Set sensitive_columns and then validate if needed."""
         self._sensitive_columns = sensitive_columns
         if self._sensitive_columns:
@@ -217,8 +216,6 @@ class PandasCompare(BaseCompare):
 
         # Cast to lowercase if applicable
         if self.cast_column_names_lower:
-            # Need to directly modify _sensitive_columns here otherwise it will
-            # infinitely recurse in @sensitive_columns.setter
             self._sensitive_columns = [col.lower() for col in self.sensitive_columns]
 
         # Check duplicates
@@ -244,7 +241,7 @@ class PandasCompare(BaseCompare):
                 "sensitive columns are already hidden, call reveal_sensitive_columns() first"
             )
 
-        self.sensitive_columns = sensitive_columns  # validates through setter
+        self._set_sensitive_columns(sensitive_columns)  # validates through private setter
         sensitive = set(self.sensitive_columns)
         common_cols = self.intersect_columns() - set(self.join_columns)
 
@@ -281,7 +278,7 @@ class PandasCompare(BaseCompare):
             return None
 
         LOG.debug("Revealing sensitive columns and re-comparing dfs")
-        self.sensitive_columns = None
+        self._set_sensitive_columns(None)
         self.column_stats.clear()
         self._compare(ignore_spaces=self.ignore_spaces, ignore_case=self.ignore_case)
 
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -249,21 +249,23 @@ class PandasCompare(BaseCompare):
 
         # Hide columns in unq_rows
         for df in (self.df1_unq_rows, self.df2_unq_rows):
-            LOG.debug("Hiding sensitive columns in unq_rows")
-            cols_to_hash = [col for col in df.columns if col in sensitive]
-            for col in cols_to_hash:
+            LOG.debug(f"Hiding sensitive columns in unq_rows: {df.columns}")
+            cols_to_hide = [col for col in df.columns if col in sensitive]
+            for col in cols_to_hide:
                 df[col] = "*******"
 
         # Hide columns in intersect_rows
-        LOG.debug("Hiding sensitive columns in intersect_rows")
-        cols_to_hash = [
+        LOG.debug(
+            f"Hiding sensitive columns in intersect_rows: {self.intersect_rows.columns}"
+        )
+        cols_to_hide = [
             *[col for col in self.join_columns if col in sensitive],
             *[col for col in self.df1_unq_columns() if col in sensitive],
             *[col for col in self.df2_unq_columns() if col in sensitive],
             *[f"{col}_{self.df1_name}" for col in common_cols if col in sensitive],
             *[f"{col}_{self.df2_name}" for col in common_cols if col in sensitive],
         ]
-        for col in cols_to_hash:
+        for col in cols_to_hide:
             self.intersect_rows[col] = "*******"
 
     def reveal_sensitive_columns(self) -> None:
--- a/datacompy/pandas.py
+++ b/datacompy/pandas.py
@@ -258,16 +258,15 @@ class PandasCompare(BaseCompare):
         )
 
         # Hide columns in unq_rows
-        for df in (self.df1_unq_rows, self.df2_unq_rows):
-            LOG.debug(f"Hiding sensitive columns in unq_rows: {df.columns}")
+        for df_name in ("df1_unq_rows", "df2_unq_rows"):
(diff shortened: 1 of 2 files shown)


datacompy/pandas.py
```diff
@@ -233,40 +226,46 @@ def _validate_sensitive_columns(self) -> None:
...
-        for df in (self.df1, self.df2):
-            # Process only the columns that actually exist in the current dataframe
-            cols_to_hash = [c for c in self.sensitive_columns if c in df.columns]
+        self.sensitive_columns = sensitive_columns  # validates through setter
+        sensitive = set(self.sensitive_columns)
+        common_cols = self.intersect_columns() - set(self.join_columns)
 
+        # Hide columns in unq_rows
+        for df in (self.df1_unq_rows, self.df2_unq_rows):
+            LOG.debug(f"Hiding sensitive columns in {df}")
```
fdosani: Might want to change `df` to something like `df.columns`. I think this will print out the entire dataframe.

datacompy/pandas.py
```diff
@@ -233,40 +226,46 @@ def _validate_sensitive_columns(self) -> None:
...
-    def _hash_sensitive_columns(self) -> None:
-        """Hash sensitive columns in both dataframes."""
-        if not self.sensitive_columns:
-            return None
-
-        # Logs warning only if sensitive columns is set and validation passes
-        LOG.warning(
-            "dataframes with columns in sensitive_columns will be modified inplace."
-        )
+    def hide_sensitive_columns(self, sensitive_columns: List[str]) -> None:
```
fdosani, line 241: Should we add some validation or logging if the column we are asking to hide actually exists?

datacompy/pandas.py
```diff
@@ -233,40 +226,46 @@ def _validate_sensitive_columns(self) -> None:
...
+        if self.sensitive_columns is None:
+            return None
 
-                # Ensure all nulls are consistently represented as pd.NA in the hashed column
-                if is_null.any():
-                    df.loc[is_null, col] = pd.NA
+        LOG.debug("Revealing sensitive columns and re-comparing dfs")
+        self.sensitive_columns = None
+        self.column_stats.clear()
+        self._compare(ignore_spaces=self.ignore_spaces, ignore_case=self.ignore_case)
```
fdosani, line 292: We should add something in the documentation about this: Something like:

> "This re-runs the full comparison to restore original values"

datacompy/pandas.py
```diff
@@ -233,40 +226,46 @@ def _validate_sensitive_columns(self) -> None:
...
-            return None
-
-        # Logs warning only if sensitive columns is set and validation passes
-        LOG.warning(
-            "dataframes with columns in sensitive_columns will be modified inplace."
-        )
+    def hide_sensitive_columns(self, sensitive_columns: List[str]) -> None:
+        """Hide sensitive columns of df1 or df2 if applicable in the compare."""
+        # Don't allow hiding columns again before first revealing
+        if self.sensitive_columns:
```
fdosani, line 244: Do we want to be consistent here with `reveal_sensitive_columns`. We check for `if self.sensitive_columns is None:` and then just `if self.sensitive_columns:`

tests/test_pandas.py
```diff
@@ -2173,98 +2174,222 @@ def test_array_comparator_pandas():
...
+    )
+    assert (
+        compare.sample_mismatch("a").sort_values("a").reset_index(drop=True).loc[1, "a"]
+        == 2
+    )
     # Just render the report to make sure it renders.
     compare.report()
 
 
 def test_sensitive_columns_missing():
```
fdosani, line 2360: wondering if we can include a test where `sensitive_columns` are not in either dataframe. I guess for now it might silently do nothing, but it should log some warning message.

fdosani (changes requested):
Initial thoughts on the review. Nice work!

datacompy/pandas.py
sh13m: I think just logging a warning if the sensitive column does not exist in both dataframes is sufficient, will add a test for this as well.

datacompy/pandas.py
sh13m: By consistency, do you mean if we want both of the checks to be the same, or that the `is None` condition doesn't handle empty lists while the other `if self.sensitive_columns:` condition implicitly handles both? 

I think having the return/error condition at the top before the main function body makes sense, but I would agree that there is some ambiguity between how `None` and `[]` inputs are handled. We can change the `reveal_sensitive_columns` condition to `if not self.sensitive_columns:`. This way we functionally treat `None` and `[]` as the same.

sh13m:
Not sure why tests were randomly failing, might have been a github issue.

datacompy/pandas.py
```diff
@@ -224,49 +216,75 @@ def _validate_sensitive_columns(self) -> None:
...
         # Check duplicates
         duplicates = {c for c, n in Counter(self.sensitive_columns).items() if n > 1}
         if duplicates:
             raise ValueError(f"duplicate columns: {duplicates}")
 
-    def _hash_sensitive_columns(self) -> None:
-        """Hash sensitive columns in both dataframes."""
-        if not self.sensitive_columns:
-            return None
+        # Warn if column not in both df1 and df2
```
```suggestion
        # Warn if column not found in either dataframe
```

datacompy/pandas.py
```diff
@@ -224,49 +216,75 @@ def _validate_sensitive_columns(self) -> None:
...
+        unused = [
+            col
+            for col in self.sensitive_columns
+            if (col not in self.df1.columns) and (col not in self.df2.columns)
+        ]
+        if unused:
+            LOG.warning(
+                f"sensitive columns not found in both df1 and df2 will be ignored: {unused}"
```
```suggestion
                f"sensitive columns not found in either df1 or df2 will be ignored: {unused}"
```

datacompy/pandas.py
```diff
@@ -210,8 +203,7 @@ def sensitive_columns(self) -> List[str] | None:
         """Get the list of sensitive columns."""
         return self._sensitive_columns
 
-    @sensitive_columns.setter
-    def sensitive_columns(self, sensitive_columns: List[str] | None) -> None:
+    def _set_sensitive_columns(self, sensitive_columns: List[str] | None) -> None:
```
fdosani: Might have missed this, but we had `@sensitive_columns.setter` and moved away from it?

tests/test_pandas.py
```diff
@@ -2173,132 +2174,357 @@ def test_array_comparator_pandas():
...
+    assert compare.intersect_rows.reset_index(drop=True).loc[0, "b_match"]
     # Just render the report to make sure it renders.
     compare.report()
 
 
 def test_sensitive_columns_setter():
     df1 = pd.DataFrame([{"a": 1, "b": 2}])
     df2 = pd.DataFrame([{"a": 1, "b": 2}])
-    compare = datacompy.PandasCompare(df1, df2, join_columns=["a"])
+    compare = PandasCompare(df1, df2, join_columns=["a"])
```
fdosani:
```suggestion
    compare = datacompy.PandasCompare(df1, df2, join_columns=["a"])
```

datacompy/pandas.py
fdosani: Thinking about this a bit more. How about be combine set and validation into something like? 

```python
    def _set_and_validate_sensitive_columns(
        self, sensitive_columns: List[str] | None
    ) -> None:
        """Set and validate sensitive columns.

        Normalizes empty lists to None so there is only one representation
        for "no sensitive columns".
        """
        self._sensitive_columns = sensitive_columns or None
        if not self._sensitive_columns:
            return

        if not all(isinstance(c, str) for c in self._sensitive_columns):
            raise TypeError("sensitive_columns must be a list of strings")

        # Cast to lowercase if applicable
        if self.cast_column_names_lower:
            self._sensitive_columns = [col.lower() for col in self._sensitive_columns]

        # Check duplicates
        duplicates = {c for c, n in Counter(self._sensitive_columns).items() if n > 1}
        if duplicates:
            raise ValueError(f"duplicate columns: {duplicates}")

        # Warn if column not in either df1 or df2
        unused = [
            col
            for col in self._sensitive_columns
            if (col not in self.df1.columns) and (col not in self.df2.columns)
        ]
        if unused:
            LOG.warning(
                f"sensitive columns not found in either df1 or df2 will be ignored: {unused}"
            )

    def hide_sensitive_columns(self, sensitive_columns: List[str]) -> None:
        """Hides sensitive columns of df1 or df2 if applicable in the compare."""
        # Don't allow hiding columns again before first revealing
        if self.sensitive_columns:
            raise ValueError(
                "sensitive columns are already hidden, call reveal_sensitive_columns() first"
            )

        self._set_and_validate_sensitive_columns(sensitive_columns)
        if not self.sensitive_columns:
            return

        sensitive = set(self.sensitive_columns)
        sensitive_with_suffixes = (
            sensitive
            | {f"{c}_{self.df1_name}" for c in sensitive}
            | {f"{c}_{self.df2_name}" for c in sensitive}
        )

        # Hide columns in unq_rows
        for df in (self.df1_unq_rows, self.df2_unq_rows):
            cols_to_hide = [col for col in df.columns if col in sensitive]
            for col in cols_to_hide:
                df[col] = "*******"

        # Hide columns in intersect_rows
        cols_to_hide = [
            col
            for col in self.intersect_rows.columns
            if col in sensitive_with_suffixes
        ]
        for col in cols_to_hide:
            self.intersect_rows[col] = "*******"
```

fdosani (commented):
Some comments and thoughts, appreciate the effort and improve on the hashing!

datacompy/pandas.py
sh13m: Yeah, recent change since I don't think we'd want to have users modifying the list directly and causing undefined behaviour. I'm good with combining set and validate as well, will make the changes soon.

sh13m: nvm I am dumb lmao ignore this

fdosani (approved):

PR #500, opened 2026-03-11
Release v0.19.4

### LINKED ISSUE #499: Snowflake: TypeError: must be real number, not NoneType
snowflake compare output where the mismatch stats has a max_diff of None leading to this jinja template error: TypeError: must be real number, not NoneType. 

Seems to a missing condition for the max_diff calculation:

Line 512 in `snowflake.py` would become:
```python
max_diff = max_diff.result()[0][0]
if max_diff is None:
    max_diff = 0
```

Needs to be fixed in `0.19.4 `and `1.0.0b2`

fdosani:
Release v0.19.4
----------------------
Fixes #499 
bug fix: Snowflake compare when max_diff is None set to 0.

### DIFF
--- a/datacompy/__init__.py
+++ b/datacompy/__init__.py
@@ -18,4 +18,4 @@ Originally started to be something of a replacement for SAS's PROC COMPARE for P
 Then extended to carry that functionality over to Spark Dataframes.
 """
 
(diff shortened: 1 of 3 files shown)


rhaffar (approved):
LGTM

PR #501, opened 2026-03-12
fix: Snowflake compare when max_diff is None set to 0.

fdosani:
v1.0.0 fix for #499


rhaffar (approved):
LGTM

rhaffar (commented):
btw I've noticed similar errors in other compares - may be something we address at the templating level as well?

fdosani:
> btw I've noticed similar errors in other compares - may be something we address at the templating level as well?

Which ones specifically? Maybe we can create an issue to track this with some examples?

PR #502, opened 2026-03-13
feat: hide sensitive columns in PolarsCompare

sh13m:
Implements `sensitive_columns` in a similar fashion to https://github.com/capitalone/datacompy/pull/494 but done in polars instead.

- Slight difference in `intersect_rows`. In pandas, the merge does not suffix the dataframe name for columns that are only in one of the dataframes, while the way merging is implemented in PolarsCompare always adds the suffix. 
- Refactor: moved `_set_and_validate_sensitive_columns`, the `sensitive_columns` property, and `reveal_sensitive_columns` into `BaseCompare` since the implementations are identical and should remain so for the other compares.

### DIFF
--- a/datacompy/base.py
+++ b/datacompy/base.py
@@ -23,8 +23,9 @@ two dataframes.
 
 import logging
 from abc import ABC, abstractmethod
+from collections import Counter
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 from ordered_set import OrderedSet
@@ -56,6 +57,46 @@ class BaseCompare(ABC):
     def df2(self, df2: Any) -> None:
         """Check that it is a dataframe and has the join columns."""
         pass
+
+    @property
+    def sensitive_columns(self) -> List[str] | None:
+        """Get the list of sensitive columns."""
+        return self._sensitive_columns
+
+    def _set_and_validate_sensitive_columns(
+        self, sensitive_columns: List[str] | None
+    ) -> None:
+        """Set and validate sensitive columns.
+
+        Normalizes empty lists to None so there is only one representation
+        for "no sensitive columns".
+        """
+        self._sensitive_columns = sensitive_columns or None
+        if not self._sensitive_columns:
+            return
+
+        if not all(isinstance(c, str) for c in self.sensitive_columns):
+            raise TypeError("sensitive_columns must be a list of strings")
+
+        # Cast to lowercase if applicable
+        if self.cast_column_names_lower:
+            self._sensitive_columns = [col.lower() for col in self.sensitive_columns]
+
+        # Check duplicates
+        duplicates = {c for c, n in Counter(self.sensitive_columns).items() if n > 1}
+        if duplicates:
+            raise ValueError(f"duplicate columns: {duplicates}")
+
+        # Warn if column not found in either dataframe
+        unused = [
+            col
+            for col in self.sensitive_columns
+            if (col not in self.df1.columns) and (col not in self.df2.columns)
+        ]
+        if unused:
+            LOG.warning(
+                f"sensitive columns not found in either df1 or df2 will be ignored: {unused}"
+            )
 
     @abstractmethod
     def _validate_dataframe(
@@ -181,6 +222,24 @@ class BaseCompare(ABC):
             The report, formatted according to the template.
         """
         pass
+
+    def reveal_sensitive_columns(self) -> None:
+        """Reveals all sensitive columns.
+
+        Notes
+        -----
+        - This re-runs the full comparison to restore original values.
+        - Revealing sensitive columns when there aren't any is treated as a NOP
+          to avoid redundant computations.
+        """
+        # Don't do anything if there aren't any sensitive columns
+        if not self.sensitive_columns:
+            return
+
+        LOG.debug("Revealing sensitive columns and re-comparing dfs")
+        self._set_and_validate_sensitive_columns(None)
+        self.column_stats.clear()
+        self._compare(ignore_spaces=self.ignore_spaces, ignore_case=self.ignore_case)
 
     def only_join_columns(self) -> bool:
         """Boolean on if the only columns are the join columns."""
--- a/datacompy/base.py
+++ b/datacompy/base.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 """
-Compare two Pandas DataFrames.
+Base module for comparing two DataFrames.
 
 Originally this package was meant to provide similar functionality to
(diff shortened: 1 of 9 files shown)


sh13m:
will set to ready once the other PandasCompare changes are merged into develop.

datacompy/polars.py
```diff
@@ -184,6 +186,107 @@ def df2(self, df2: pl.DataFrame) -> None:
             "df2", cast_column_names_lower=self.cast_column_names_lower
         )
 
+    @property
+    def sensitive_columns(self) -> List[str] | None:
+        """Get the list of sensitive columns."""
+        return self._sensitive_columns
+
+    def _set_and_validate_sensitive_columns(
```
fdosani: `_set_and_validate_sensitive_columns`, the `sensitive_columns` property, and `reveal_sensitive_columns` are identical to the `PandasCompare` implementations. Worth considering pulling these into `BaseCompare` in a follow-up, leaving only `hide_sensitive_columns` in each subclass since that's the only method with backend-specific logic.

tests/test_polars.py
```diff
@@ -2287,3 +2288,353 @@ def test_columns_with_mismatches_multiple_join_columns():
...
+    assert not isinstance(compare.df1[0, "c"], str)
+    assert len(compare.df1_unq_rows) == 0
+    assert compare.intersect_rows[0, "b_df1"] == "*******"
+    assert compare.intersect_rows[0, "b_df2"] == "*******"
+    assert compare.intersect_rows[0, "c_df1"] == "*******"
+    assert compare.intersect_rows[0, "c_df2"] == "*******"
+
+
+def test_sensitive_columns_numeric_types_with_tolerance():
+    """Verify that hashing works for different numeric types with tolerance."""
```
fdosani:
```suggestion
    """Verify that hiding works for different numeric types with tolerance."""
```

tests/test_polars.py
```diff
@@ -2287,3 +2288,353 @@ def test_columns_with_mismatches_multiple_join_columns():
...
+    df2 = pl.DataFrame([{"a": 1, "b": 2}])
+
+    compare = PolarsCompare(df1, df2, join_columns=["a"])
+    # Duplicate columns should raise ValueError during hide_sensitive_columns()
+    with pytest.raises(ValueError, match=r"duplicate columns: {'b'}"):
+        compare.hide_sensitive_columns(["b", "b"])
+
+
+def test_sensitive_columns_numeric_types():
+    """Verify that hashing works for different numeric types without LossySetitemError."""
```
fdosani:
```suggestion
    """Verify that hiding works for different numeric types without LossySetitemError."""
```

datacompy/polars.py
```diff
@@ -184,6 +186,107 @@ def df2(self, df2: pl.DataFrame) -> None:
...
+        for df_name in ("df1_unq_rows", "df2_unq_rows"):
+            df = getattr(self, df_name)
+            LOG.debug(f"Hiding sensitive columns in {df_name}")
+            cols_to_hide = [col for col in df.columns if col in sensitive]
+            if not cols_to_hide:  # skip if empty
+                continue
+            setattr(
+                self,
+                df_name,
+                df.with_columns([pl.lit("*******").alias(col) for col in cols_to_hide]),
```
fdosani, line 217: Question/Validation check:`pl.lit("*******") `will change the column dtype from its original type (e.g., `Int64`, `Float64`) to `String`. This is fine for display purposes but differs from the Pandas behaviour where the column dtype is also changed (to `object`). 

Just confirming this is the intended behaviour the `dtype` change in `unq_rows` and `intersect_rows` after hiding could potentially affect downstream code that inspects `dtypes`. Not a blocker since `column_stats` captures `dtypes` from df1/df2 (which are untouched), not from the result dataframes.

fdosani (commented):
Thanks for the PR, nice work porting this over to Polars!
Left a few minor comments

datacompy/polars.py
sh13m: Yeah I think that's a good idea. Planning to add column hiding to the other types of compares soon, so I can move the methods to base sometime then (assuming nothing weird shows up which would require a different implementation, probably won't).

datacompy/polars.py
sh13m: I guess this is more of a side effect of the implementation since we are modifying `unq_rows` and `intersect_rows` by doing column replacements with strings of `*******`. The difference in Pandas behaviour is just because text data types are stored with the NumPy `object` dtype. 

The only case where I can think of downstream code being affected is if users are trying to call methods to do further analysis on a compare which has been hidden. In which case there are a few ideas I have in mind:
- Have a note/warning of this behaviour documented somewhere. 
- Only rely on `column_stats` when inspecting `dtypes`.
- Don't allow downstream code to be executed if columns are hidden and raise and error. 

The second solution makes the most sense to me.

datacompy/polars.py
fdosani: Did you want to make that shift now so we can review discuss it here?

sh13m: @fdosani changes made in latest commits, let me know what you think.

datacompy/base.py
fdosani, line 38: Might need this here or `SparkSQLCompare`, `SnowflakeCompare` will crash if someone tries to access: `sensitive_columns`

```suggestion

      _sensitive_columns: List[str] | None = None
```

datacompy/polars.py
```diff
@@ -113,6 +113,7 @@ def __init__(
     ) -> None:
         self.cast_column_names_lower = cast_column_names_lower
         self.custom_comparators = custom_comparators or []
+        self._set_and_validate_sensitive_columns(None)
```
fdosani: Remove this in favour of: https://github.com/capitalone/datacompy/pull/502/changes#r3019500651 ?

datacompy/pandas.py
fdosani: Remove this in favour of: https://github.com/capitalone/datacompy/pull/502/changes#r3019500651 ?

fdosani (commented):
We should add a test to check that `SparkSQLCompare` and `SnowflakeCompare` don't crash when accessing `.sensitive_columns` after construction.

Aside from that just the comment on adding `_sensitive_columns` as a class level attribute especially for the init of it.

datacompy/base.py
sh13m: Planning to port over `sensitive_columns` to both soon, correct me if I'm wrong but I assume at that point we wouldn't need this here? I can add this + the tests in for now and remove them later in their respective PRs if we want.

fdosani: Sorry for the delay! Been swamped! Sure just add them in for now so things don't crash.

sh13m: Haha no worries, sure I'll add them in.

fdosani (approved):
LGTM. Thanks for sticking with this long review.

Issue #503, opened 2026-03-16
Snowflake: dependency using deprecated pkg_resources module


sh13m:
I believe we currently have `snowflake-snowpark-python>=1.26` in the requirements (dev branch, unpinned in main actually). The `pkg_resources` module has been very recently removed in `setuptools==82.0.0` (open issue [here](https://github.com/pypa/setuptools/issues/5174)), which breaks the snowflake imports in older versions. 
- In `snowflake-snowpark-python`, `pkg_resources` was deprecrated in [v1.37.0](https://github.com/snowflakedb/snowpark-python/releases/tag/v1.37.0).
- So we would either want to pin `snowflake-snowpark-python>=1.37.0`, or `setuptools<=81.0.0`  to fix this issue.
  - More leaning towards `snowflake-snowpark-python>=1.37.0` since `pkg_resources` has been deprecated for a while.
  - Feel like pinning `setuptools<=81.0.0` might cause more issues down the line.

PR #505, opened 2026-03-30
chore: v1 bump snowflake-snowpark-python minimum to 1.37

sh13m:
Should fix https://github.com/capitalone/datacompy/issues/503 for the develop branch.
- bumps optional `snowflake-snowpark-python`  to `1.37` minimum which is the version that deprecates `pkg_resources`.


sh13m:
@fdosani let me know what you think, I think we should also pin snowflake for `v0.19`.

fdosani (approved):
LGTM!

PR #510, opened 2026-04-01
feat: hide sensitive columns in SparkSQLCompare

sh13m:
Ported over sensitive column hiding to `SparkSQLCompare`.
- Moved some methods over into `BaseCompare` to be synced with `PolarsCompare` PR (possibly subject to change).
- Added unit tests. 
- aside: removed caching of `df1` and `df2`.

### DIFF
--- a/datacompy/spark.py
+++ b/datacompy/spark.py
@@ -447,14 +447,6 @@ class SparkSQLCompare(BaseCompare):
             {c: f"{c}_{self.df2_name}" for c in temp_join_columns}
         )
 
-        # cache only if enabled
-        if self.cache_intermediates:
-            LOG.debug("Caching intermediate dataframes")
-            df1.cache()
-            df2.cache()
-        else:
-            LOG.debug("Caching disabled - skipping cache() calls")
-
         # NULL SAFE Outer join using ON
         df1.createOrReplaceTempView("df1")
         df2.createOrReplaceTempView("df2")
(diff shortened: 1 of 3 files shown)


fdosani:
@sh13m you might need to rebase and resolve some conflicts before reviewing here. Same for the Snowflake one.

datacompy/spark.py
```diff
@@ -221,6 +221,52 @@ def df2(self, df2: "pyspark.sql.DataFrame") -> None:
...
+        cols_to_hide = [
+            col for col in self.intersect_rows.columns if col in sensitive_with_suffixes
+        ]
+        if not cols_to_hide:  # skip if empty
+            return
+        select_cols = [
+            F.lit("*******").alias(col) if col in cols_to_hide else F.col(col)
+            for col in self.intersect_rows.columns
+        ]
+        self.intersect_rows = self.intersect_rows.select(select_cols)
```
fdosani, line 271:
```suggestion
        if self.cache_intermediates:
            self.intersect_rows.unpersist()
        self.intersect_rows = self.intersect_rows.select(select_cols)
```
The old intersect_rows is cached, `unpersist` it before replacing to avoid orphaned cached data (including unmasked sensitive values) in executor memory.

datacompy/spark.py
fdosani: We should delete this section if possible. Looking into this a bit more, it might not make sense and just lead to move overhead.

Remove: 

```python
  if self.cache_intermediates:
      LOG.debug("Caching intermediate dataframes")
      df1.cache()
      df2.cache()
  else:
      LOG.debug("Caching disabled - skipping cache() calls")
```

fdosani (changes requested):
2 small edits. Otherwise LGTM!

fdosani (approved):
LGTM, thanks for the updates

PR #519, opened 2026-05-28
fix: make Spark backend compatible with spark.sql.ansi.enabled=true (Spark 4.x default)

### LINKED ISSUE #418: Validate and add support for Spark 4.0.0
Spark 4.0.0 is out now. Should spend some time to understand any differences and test if we are compatible with the latest version: https://spark.apache.org/docs/latest/sql-migration-guide.html#upgrading-from-spark-sql-35-to-40

fdosani:
Fixes #418 

- Spark 4.x defaults `spark.sql.ansi.enabled `to `true`, breaking the Spark backend which was written assuming false
- Replaced `CAST` with `TRY_CAST` in `calculate_max_diff()` so non-castable values return NULL instead of throwing
- Fixed cross-type string equality in `_generate_id_within_group()` to avoid `AnalysisException` on non-string join columns
- Cast integer columns to double before arithmetic in `SparkNumericComparator` to prevent ANSI-mode overflow
- Replaced `NaN` correction logic with a readable flat when chain that is also symmetric for both `NaN` positions
- Fixed mixed date/string comparisons in `SparkStringComparator` to use T`RY_CAST(string AS date)`: preserves truncation semantics and handles malformed input without throwing

### DIFF
--- a/datacompy/comparator/numeric.py
+++ b/datacompy/comparator/numeric.py
@@ -308,13 +308,17 @@ class SparkNumericComparator(BaseComparator):
                 nan_col1 = psf.isnan(col1_expr) if col1_can_be_nan else psf.lit(False)
                 nan_col2 = psf.isnan(col2_expr) if col2_can_be_nan else psf.lit(False)
                 return (
-                    psf.when(nan_col1 & nan_col2, psf.lit(True))   # NaN == NaN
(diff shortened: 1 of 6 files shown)


star11293:
Pulled the branch and ran the full suite locally on PySpark 4.1.1, all 493 tests pass including the new comparator tests for non numeric strings.

ak-gupta (approved):

PR #523, opened 2026-06-04
fix: update Spark versions and add ANSI mode tests for pytest

fdosani:
Added ANSI mode pytest steps to both Spark CI jobs in `test-package.yml` using `pytest-ansi.ini`.


rhaffar (approved):
LGTM

PR #532, opened 2026-06-18
feat: add programmatic Report API with structured comparison data

### LINKED ISSUE #531: feat: modular reporting framework
## Summary

Each DataComPy backend duplicates ~150 lines of `_get_*` helper methods that build dict structures feeding a shared Jinja2 template. The contract between backends and template is enforced only by copy-paste convention, and small inconsistencies have crept in over time.

This issue proposes replacing that duplication with a shared typed reporting framework and a public programmatic API.

## Proposed changes

- **`datacompy/report.py`** (new) — frozen dataclasses (`ReportData`, `ColumnSummary`, `RowSummary`, `ColumnComparison`, `MismatchStat`, `MismatchStats`, `UniqueRowsData`) as the single source of truth for the template contract. `ReportData` should own rendering via `render()`, `to_html()`, `save()`, and `to_dict()`.
- **`datacompy/base.py`** — add a concrete `build_report_data()` method shared across all backends, replacing the 20 duplicated helpers. Backends should override only small primitives where needed (`_table_shape`, `_row_count`, `_select_first_n_columns`).
- **All backends** — remove `_get_*` helpers and `report()` body. Spark/Snowflake should cache `.count()` calls to avoid redundant distributed actions.
- **Cleanups** — mismatch stats should be sorted consistently across all backends; dead `columns` sub-key removed; Spark's leaked outer keys dropped.
- **New tests** — unit tests for dataclasses and rendering; snapshot regression tests for Pandas and Polars report output.
- **Docs** — new Report API page, README programmatic access section.

## New public API

All new dataclasses and `build_report_data()` should be exported as public API, enabling programmatic access to comparison results without rendering:

```python
data = compare.build_report_data()
data.row_summary.unequal_rows      # structured access
data.to_dict()                     # JSON-serializable
data.render()                      # same text as compare.report()
data.save("report.html")           # HTML export

### PR DESCRIPTION
**Pull request by fdosani**
Closes #531 
  
- Introduces a new `ReportData` dataclass in `datacompy/report.py` that exposes structured comparison results (mismatches, unmatched rows, column stats) as a programmatic API, accessible via a new `.report` property on all compare classes
- Consolidates report generation logic into `datacompy/base.py`, removing duplicated summary methods from `pandas.py`, `polars.py`, `snowflake.py`, and `spark.py`
- Adds snapshot regression tests for Pandas and Polars report output to catch unintended formatting changes
- Adds `docs/source/report_api.rst` documenting the new API and updates the template guide.

### CODE REVIEW

**Comment by fdosani** [comment]
@ak-gupta @rhaffar worth having more of a discussion on this one, as its a bit involved on the backend.

File: datacompy/polars.py [hunk 1]
```diff
@@ -703,246 +700,6 @@ def all_mismatch(self, ignore_matching_cols: bool = False) -> pl.DataFrame:
             .select(self.join_columns + return_list)
         )
```
rhaffar, line 705: I'm guessing the polars compare doesn't implement `_select_first_n_columns` and `_column_names` since it's using the default implementation? If all the other compare classes have to overwrite it anyways, might opt to just make the base class method abstract and force all compare classes to implement them.

datacompy/base.py
```diff
@@ -191,7 +194,199 @@ def all_mismatch(self, ignore_matching_cols: bool = False) -> Any:
...
+                    )
+                )
+                if col["unequal_cnt"] > 0:
+                    sample_list.append(
+                        self.sample_mismatch(
+                            col["column"], sample_count, for_display=True
+                        )
+                    )
+
+        if any_mismatch:
```
rhaffar: If I'm not mistaken, you can just check if the stat_list is empty instead of keeping track of this flag

rhaffar (commented):
I like the centralizing of our reporting logic here - it's so similar between the different backends, I think keeping that intent in one spot makes more sense than re-implementing it for each backend. No real gripes from me here on any of these changes - just a few small comments.

datacompy/base.py
fdosani: Fixed! Thanks

datacompy/polars.py
fdosani: I think I'd like to leave this as is for now. Maybe we can have an offline discussion about this? Just want to avoid duplication here as much as possible.

rhaffar (approved):
LGTM

PR #534, opened 2026-06-25
feat: add datacompy compare CLI with multi-backend support

### LINKED ISSUE #530: feat: Command Line Interface
## Summary

DataComPy is currently a Python library only — users must write a script to compare two datasets. A command-line interface would make it immediately accessible for ad-hoc data validation, CI pipelines, and use by non-Python practitioners without any boilerplate.

## Motivation

CICD pipelines often run as shell tasks (Airflow `BashOperator`, GitHub Actions, GitLab CI) where Python entry points are awkward. A native CLI that exits non-zero on mismatch and emits machine-readable JSON is the highest-leverage single CICD improvement DataComPy could add.

Other common friction points:

- **Ops / data engineers** want to diff two CSV or Parquet files without opening a Python REPL
- **Analysts** want a one-liner to spot-check a file diff without writing a throw-away script every time
- **Automation** wants to shell out to `datacompy compare ...` and capture a structured result

## Proposed interface

```bash
# Basic comparison
datacompy compare \
    --left s3://bucket/before.parquet \
    --right s3://bucket/after.parquet \
    --on id,date \
    --backend pandas \
    --max-unequal-rows 0 \
    --json
# exit 0 on match, 1 on mismatch, 2 on error

# Local files, HTML report written to file
datacompy compare \
    --left df1.csv \
    --right df2.csv \
    --on id \
    --output report.html

# Tolerances
datacompy compare \
    --left before.parquet \
    --right after.parquet \
    --on account_id \
    --abs-tol 0.01 \
    --rel-tol 0.001

# Ignore whitespace and case
datacompy compare \
    --left df1.csv \
    --right df2.csv \
    --on id \
    --ignore-spaces \
    --ignore-case

# Control report verbosity
datacompy compare \
    --left df1.csv \
    --right df2.csv \
    --on id \
    --sample-count 20 \
    --column-count 5

### PR DESCRIPTION
**Pull request by fdosani**
Implements the `datacompy compare` CLI (closes #530), enabling dataset comparisons from the shell and CI/CD pipelines without writing Python.

- Adds `datacompy/cli/` subpackage: `parser.py`, `compare.py`, `backends.py`, `loaders.py`, `sessions.py`, `output.py`, `errors.py`
- Supports all four backends: polars (default), pandas, spark, snowflake
- Loads CSV, Parquet, and JSON files; Snowflake can compare native table refs (`db.schema.table`) or stage local files to a temp table
- Exit codes: `0` match, `1` mismatch/threshold violated, `2` error
- `--max-unequal-rows N` for threshold-based CI assertions
- `--json` for machine-readable output; `--quiet` for exit-code-only pipelines
- `--debug` re-raises unexpected exceptions for bug reporting
- Argument validation split between argparse `type=` callables (value rules) and `_validate_arg_combinations` (cross-argument rules)
- Session lifecycle managed via `contextlib.ExitStack` — Spark and Snowflake sessions close on both success and failure
- All optional backend imports are deferred so `datacompy --version` does not pay the pyspark/snowflake import cost

### CODE REVIEW

File: datacompy/cli/backends.py [hunk 1]
```diff
@@ -0,0 +1,204 @@
...
+        df2,
+        join_columns=args.on,
+        abs_tol=args.abs_tol,
+        rel_tol=args.rel_tol,
+        df1_name=args.df1_name,
+        df2_name=args.df2_name,
+        ignore_spaces=args.ignore_spaces,
+        ignore_case=args.ignore_case,
+        cast_column_names_lower=args.cast_column_names_lower,
+    )
```
rhaffar: We could probably keep this to a single return call by making on_index false if args.on isn't set.

datacompy/cli/compare.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    if args.on_index and args.backend != "pandas":
+        raise BadArgsError("--on-index is only supported with --backend pandas.")
+    if args.on_index and args.on:
+        raise BadArgsError("--on and --on-index are mutually exclusive.")
+    if not args.on_index and not args.on:
+        raise BadArgsError(
+            "--on is required (or --on-index for the pandas backend). "
+            "Specify at least one join column with --on COL."
+        )
```
rhaffar, line 87: Right now we do this check in the Pandas Comparator - if we just pass on_index and join_columns directly to the comparator, I think we could get rid of these checks.

datacompy/cli/loaders.py
```diff
@@ -0,0 +1,337 @@
...
+        When a ref cannot be resolved or a local file cannot be staged.
+    LoadError
+        On file read or Snowflake staging errors.
+    """
+    if is_snowflake_ref(ref):
+        return _expand_table_ref(session, ref)
+    return _stage_file_to_snowflake(session, ref, fmt, csv_delimiter)
+
+
+def _stage_file_to_snowflake(
```
rhaffar: I think the use case for comparing local files is very limited for Snowflake - that said Snowpark has a set of file readers that can be used to read files from a Snowflake internal stage. We could use that to read in Snowpark dataframes that we can pass directly to the Snowflake compare object?

datacompy/cli/parser.py
```diff
@@ -0,0 +1,278 @@
...
+    # ---- join keys -------------------------------------------------------------
+    keys = cmp.add_argument_group("join keys")
+    keys.add_argument(
+        "--on",
+        action="append",
+        dest="on",
+        default=None,
+        metavar="COL",
+        help="Join column name (required unless --on-index is used).  Repeat for composite keys: --on id --on date.",
+    )
```
rhaffar, line 156: Moreso on the side of personal preference, but could we have this just accept either a single or comma separated list of args?

rhaffar (commented):
Really nice feature Faisal - a few comments on my part. 

This would be a bigger change than the ones left in my review, but any thoughts on making compare args as passthroughs that we don't explicitly define for the CLI? I like defining the args explicitly, but I'm thinking this will have to mean that any changes to the compare interfaces will have to be reflected here from now on as well. We'd probably still need to keep a few explicit args here for special cases if we were to consider that though (like the spark name and SF config stuff).

datacompy/cli/backends.py
fdosani: Good call. I'll build the shared `kwargs` once and only add the join key that applies, so there's a single return `PandasCompare(...):`

datacompy/cli/compare.py
fdosani: Fair observation that the rules overlap with `PandasCompare`. I'd lean toward keeping them at the CLI layer though, for a couple of reasons: the comparator raises a plain `ValueError`, which isn't a `CLIError`, so `main.py` would surface it as a raw traceback instead of a friendly message + exit code 2; and the CLI message references the actual flags (`--on` / `--on-index`) rather than the library `kwarg` names. The `--on-index` is only supported with `--backend` pandas check is also CLI-only. There's a small redundancy, but it buys clean CLI errors across all four backends (polars/spark get args.on or [], so they wouldn't hit the comparator guard at all). Happy to discuss further though.

datacompy/cli/parser.py
fdosani: Sure, easy to support both. I'll keep `action="append"` and add a `type= callable` that splits each value on commas, then flatten, so `--on id,date` and `--on id` `--on date `(and a mix) all work. Only edge case is a column name that literally contains a comma, repeating `--on` stays as the escape hatch for that, and I'll note it in the help text.

datacompy/cli/loaders.py
fdosani: Agreed the `local-file → Snowflake` path is a narrow use case. On the Snowpark file readers: they're a nice fit when the data already lives on an internal/external stage, and since `SnowflakeCompare` accepts a Snowpark DataFrame directly we could pass one straight through. The wrinkle is that those readers read from a stage, not a local path, so for a genuinely local file we'd still need a `PUT` to a stage first, it relocates the upload step rather than removing it, so it's not a clear simplification over write_pandas. Two options: (a) keep `write_pandas` for local files and additionally accept `@stage/...` refs via the Snowpark reader, or (b) drop local-file support for Snowflake entirely and only accept table/stage refs. I lean toward (b) for the first release given how narrow the use case is, what's your preference?

fdosani:
Closing in favour of #544

PR #537, opened 2026-07-21
Fix import statement for SparkSQLCompare

gavincampbell:
Module structure has changed since 0.19.5


fdosani (approved):
Thanks for catching this

PR #538, opened 2026-07-24
fix: pandas boolean column comparisons

akshat62:
## Summary

Fix Pandas comparisons for Boolean columns, which were previously reported as mismatches even when corresponding values were equal.

## Root cause

The default Pandas comparator chain did not include a comparator for Boolean dtypes. Boolean columns were therefore rejected by the existing array-like, numeric, and string comparators, causing `columns_equal()` to fall back to an all-`False` result.

As a result, identical values such as `True` vs. `True` and `False` vs. `False` were incorrectly reported as unequal.

## Changes

- Add a dedicated `PandasBooleanComparator`.
- Register it in the default Pandas comparator chain before `PandasNumericComparator`.
- Compare Boolean-involved columns using exact, null-safe equality.
- Explicitly handle comparisons where either column has a Boolean dtype.
- Preserve the existing behavior of `PandasNumericComparator`.
- Add regression coverage for both `columns_equal()` and `PandasCompare`.

## Supported cases

- Native Pandas `bool`
- Nullable Pandas `boolean`
- All-null nullable Boolean columns
- Native `bool` compared with nullable `boolean`
- Boolean compared with integer values, such as:
  - `True` vs. `1`
  - `False` vs. `0`
- Boolean comparison in either column order

## Expected behavior

| Left | Right | Match |
|---|---|---|
| `True` | `True` | `True` |
| `False` | `False` | `True` |
| `True` | `False` | `False` |
| `True` | `1` | `True` |
| `False` | `0` | `True` |
| `True` | `0` | `False` |
| `pd.NA` | `pd.NA` | `True` |
| `pd.NA` | `False` | `False` |

## Testing

Regression tests cover:

- Native Boolean equality and mismatches
- Nullable Boolean columns containing `pd.NA`
- All-null nullable Boolean columns
- Native `bool` versus nullable `boolean`
- Boolean versus integer comparisons in both directions
- Ensuring non-Boolean columns are not claimed by `PandasBooleanComparator`
- Matching Boolean columns through `PandasCompare.matches()`
- Mismatch detection and generated match columns

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -40,6 +40,7 @@ This guide provides essential context for AI coding agents to be productive in t
 - **Comparison Usage**:
   ```python
   from datacompy import PandasCompare
+
   compare = PandasCompare(df1, df2, join_columns=[...])
   print(compare.report())
   ```
--- a/.github/workflows/edgetest.yml
+++ b/.github/workflows/edgetest.yml
@@ -14,19 +14,19 @@ jobs:
     runs-on: ubuntu-latest
     name: running edgetest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7
         with:
           ref: ${{ github.ref_name }}
 
       - name: Set up Python 3.12
-        uses: conda-incubator/setup-miniconda@v3
+        uses: conda-incubator/setup-miniconda@8ee1f361103df19b6f8c8655fd3967a8ecb162d5  # v4.0.1
         with:
           auto-update-conda: true
           python-version: '3.12'
           channels: conda-forge
 
       - name: Setup Java JDK
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@ad2b38190b15e4d6bdf0c97fb4fca8412226d287  # v5.3.0
         with:
           java-version: '17'
           distribution: 'adopt'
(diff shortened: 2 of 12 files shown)


fdosani (commented):
Appreciate the PR here:

One suggestion / thought before merging. The boolean logic might be better if housed in a dedicated `PandasBooleanComparator` rather than inside `PandasNumericComparator`. A few reasons:

  - Right now, a `bool` column compared against an `int` column (e.g. `True` vs `1`) still falls through to all-mismatches because the guard requires *both* columns to be bool. A dedicated comparator can claim any pair where either column is bool dtype and handle the cross-type case explicitly.

- The `datacompy/comparator/` subpackage already has `PandasNumericComparator`, `PandasStringComparator`, etc. Boolean is a distinct type and fits naturally as its own class.

The implementation would be small, the comparison logic you've already written just needs to move into a new class and be registered in `_PANDAS_DEFAULT_COMPARATORS` in `pandas.py`. 

Something like:

```python
class PandasBooleanComparator(BaseComparator):
    """Comparator for Boolean columns in Pandas."""

    def compare(self, col1: pd.Series, col2: pd.Series, **kwargs: Any) -> pd.Series | None:
        if not (pd.api.types.is_bool_dtype(col1.dtype) or pd.api.types.is_bool_dtype(col2.dtype)):
            return None
        if col1.shape != col2.shape:
            return None
        return (col1.eq(col2) | (col1.isna() & col2.isna())).fillna(False).astype(bool)
```

Happy to discuss further if helpful.

akshat62:
@fdosani Changes done as adviced. Thanks!

fdosani (dismissed):
Thanks for the PR. Just to flag this, we will need Polars, Spark, and Snowflake versions too I'd imagine to get to parity.

fdosani (dismissed):

fdosani (approved):

PR #539, opened 2026-07-25
Add Boolean comparison parity across backends

akshat62:
## Summary

Add Boolean column comparison support for Polars, PySpark, and Snowflake to bring these backends to parity with the existing Pandas Boolean comparator.

## Root cause

The default comparator chains for Polars, PySpark, and Snowflake did not include dedicated Boolean handling.

As a result, Boolean columns could fall through the existing array-like, numeric, and string comparators and be reported as mismatches even when corresponding values were equal.

## Changes

- Add dedicated Boolean comparators for:
  - `PolarsBooleanComparator`
  - `SparkBooleanComparator`
  - `SnowflakeBooleanComparator`
- Export the new comparators through `datacompy.comparator`.
- Register each Boolean comparator before the numeric comparator in its backend.
- Preserve the existing numeric comparator implementations.
- Use exact, null-safe Boolean comparison semantics.
- Avoid lossy numeric casts in Spark and Snowflake.
- Add backend-specific regression tests.

### Boolean versus Boolean

Boolean values are compared using exact, null-safe equality.

| Left | Right | Match |
|---|---|---|
| `True` | `True` | `True` |
| `False` | `False` | `True` |
| `True` | `False` | `False` |
| `None` | `None` | `True` |
| `None` | `False` | `False` |

### Boolean versus numeric

Boolean values are compared with exact numeric equivalents:

- `True` matches `1`
- `False` matches `0`
- `True` does not match `0`
- `False` does not match `1`
- aligned nulls match
- comparison works in either column order

The Spark and Snowflake implementations compare against native numeric literals rather than casting numeric columns to `double`. This preserves large integers and high-precision decimal values.

### Boolean versus unsupported types

When a Boolean column is compared with a nonnumeric, non-Boolean column, only aligned null values are considered equal.

## Backend integration

The Boolean comparators are registered before the existing numeric comparators:

```python
[
    ArrayLikeComparator(),
    BooleanComparator(),
    NumericComparator(),
    StringComparator(),
]

### DIFF
--- a/datacompy/comparator/__init__.py
+++ b/datacompy/comparator/__init__.py
@@ -21,7 +21,12 @@ from datacompy.comparator.array import (
     SnowflakeArrayLikeComparator,
     SparkArrayLikeComparator,
 )
-from datacompy.comparator.boolean import PandasBooleanComparator
+from datacompy.comparator.boolean import (
+    PandasBooleanComparator,
+    PolarsBooleanComparator,
+    SnowflakeBooleanComparator,
+    SparkBooleanComparator,
+)
 from datacompy.comparator.numeric import (
     PandasNumericComparator,
     PolarsNumericComparator,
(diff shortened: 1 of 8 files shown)


fdosani:
Just an FYI, I actually have a branch here with the same changes: https://github.com/capitalone/datacompy/tree/bool-logic

akshat62:
@fdosani Either of them can be merged. Post merge - possible to release a version ? Want to use reliable `bool` feature.

fdosani:
In favour of #540

PR #544, opened 2026-07-31
feat: add datacompy compare CLI with multi-backend support

### LINKED ISSUE #530: feat: Command Line Interface

fdosani:
Closes #530, Using this one in favour of #534 

Adds a `datacompy` console script so two datasets can be compared from a shell without writing a throwaway script, for ad hoc checks and for CI pipelines that run as shell tasks.

```bash
datacompy compare --left before.csv --right after.csv --on id
```

Exit codes: 0 match, 1 mismatch or threshold exceeded, 2 error, 130 interrupt. 
Supports the pandas, polars, Spark, and Snowflake backends over CSV, Parquet, and JSON.

Design notes:

* `parser.py` holds one declarative `OPTIONS` table that drives both the `argparse` parser and the `Compare` constructor `kwargs`. Adding a library keyword to the CLI is a single new row rather than two hand-maintained lists, and `tests/cli/test_parser.py` checks every mapping against the real constructor signature via inspect.signature, so the two cannot silently drift.
* Backend applicability, such as `--on-index ` being pandas only and `--cast-column-names-lower` not applying to Snowflake, is data on the spec row rather than a validation branch.
* `backends.py` defines a `CLIBackend` ABC mirroring the existing `BaseCompare` strategy pattern. Each backend owns its session, loaders, and constructor


datacompy/cli/backends.py
```diff
@@ -0,0 +1,448 @@
...
+        if not isinstance(raw, dict):
+            raise BadArgsError(
+                "--snowflake-config must contain a JSON object of connection "
+                f"parameters, got {type(raw).__name__}."
+            )
+        return raw
+
+    account = os.environ.get("SNOWFLAKE_ACCOUNT")
+    user = os.environ.get("SNOWFLAKE_USER")
+    password = os.environ.get("SNOWFLAKE_PASSWORD")
```
rhaffar, line 404: Given SF deprecating password auth, we might want to replace this with an (optional) check for an oauth token.

datacompy/cli/backends.py
```diff
@@ -0,0 +1,448 @@
...
+    authenticator = os.environ.get("SNOWFLAKE_AUTHENTICATOR")
+
+    missing = [
+        name
+        for name, value in (("SNOWFLAKE_ACCOUNT", account), ("SNOWFLAKE_USER", user))
```
rhaffar: If we decide to add an oauth check, user can be optional

datacompy/cli/compare.py
```diff
@@ -0,0 +1,157 @@
...
+
+    join_columns = OPTIONS_BY_FLAG["--on"].value(namespace)
+    on_index = OPTIONS_BY_FLAG["--on-index"].value(namespace)
+    if join_columns and on_index:
+        raise BadArgsError("--on and --on-index are mutually exclusive.")
+    if not join_columns and not on_index:
+        raise BadArgsError(
+            "--on is required. Specify at least one join column with --on COL, "
+            "or use --on-index with --backend pandas to join on the index."
+        )
```
rhaffar, line 113: We do check for this in the Pandas compare code - wondering if we should just passthrough both args and let the compare throw the error instead?

datacompy/cli/parser.py
```diff
@@ -0,0 +1,601 @@
...
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from functools import partial
+from pathlib import Path
+from typing import Any
+
+from datacompy.cli.errors import BadArgsError
+
+BACKEND_NAMES = ("pandas", "polars", "spark", "snowflake")
+ALL_BACKENDS = frozenset(BACKEND_NAMES)
```
rhaffar: Can these be one global constant

datacompy/cli/parser.py
```diff
@@ -0,0 +1,601 @@
...
+    """Split a single ``--on`` value on commas.
+
+    Combined with ``action="append"`` this makes ``--on id,date``,
+    ``--on id --on date`` and any mix of the two equivalent. Column names that
+    genuinely contain a comma must use the repeated form.
+    """
+    columns = [col.strip() for col in value.split(",") if col.strip()]
+    if not columns:
+        raise argparse.ArgumentTypeError("--on requires at least one column name")
+    return columns
```
rhaffar, line 133: 🔥 🚀

rhaffar (commented):
Looks good Faisal, just a few comments - thanks for incorporating some of the changes I recommended from the last PR

datacompy/cli/parser.py
fdosani: Done

datacompy/cli/parser.py
fdosani: :smile:

datacompy/cli/backends.py
fdosani: Done

datacompy/cli/backends.py
fdosani: Done

fdosani:
Thanks Raymond: 

- I added `SNOWFLAKE_TOKEN` with OAuth implied and `SNOWFLAKE_USER` optional under it (I'll check against a live accoun),
- dropped `BACKEND_NAMES` in favour of `ALL_BACKENDS`.
- On `--on`: kept the check, since `join_columns` is a required positional everywhere but pandas, without it you get an uncaught `TypeError` instead of the library's error, and only after Spark has already started a JVM and loaded both sides.

datacompy/cli/compare.py
fdosani: I think we should catch it in the CLI process since its separate from the non-CLI part of datacompy. Open to changing it if you think otherwise!

rhaffar: Makes sense, this one is more of a personal preference thing so more than ok to catch it in the CLI 👍

rhaffar (approved):
LGTM

PR #547, opened 2026-08-01
fix: size report table columns to their content

fdosani:
The table alignment was off and this fixes it. Adding in some tests, and adjusting existing ones.


rhaffar (approved):
LGTM

Issue #545, opened 2026-07-31
CLI: infer the CSV delimiter per file so .tsv works without extra flags


fdosani:
Follow-up from #544.

### Context

The CLI infers the input format per file, so mixed pairs like `--left a.csv --right b.parquet` work. The CSV *delimiter* is the exception: `--csv-delimiter` is a single value applied to both sides, defaulting to a comma.

`.tsv` was originally mapped to the `csv` format in `_EXTENSION_FORMATS`. That picked the right reader but not the right delimiter, so a `.tsv` file was recognised and then parsed with commas into one mangled column. The failure surfaced as `df1 must have all columns from join_columns: {'id'}`, which points nowhere near the real cause.

#544 removed `.tsv` from the extension table so the failure is at least honest:

```
$ datacompy compare --left a.tsv --right b.tsv --on id
datacompy: cannot infer the format of 'a.tsv' from its extension '.tsv'.
Pass --input-format csv|parquet|json.
```

Tab separated files still work with `--input-format csv --csv-delimiter '\t'`, as long as both sides use the same delimiter.

### What to do

Infer the delimiter per file from its extension, with an explicit `--csv-delimiter` overriding both sides:

```python
_DELIMITER_EXTENSIONS = {".tsv": "\t"}

def infer_delimiter(ref: str, override: str | None) -> str:
    """Return the field delimiter for *ref*, inferred from its extension."""
    if override is not None:
        return override
    return _DELIMITER_EXTENSIONS.get(Path(ref).suffix.lower(), ",")
```

This needs `--csv-delimiter`'s `Opt.default` changed from `","` to `None` so that "absent" stays distinguishable from "explicitly set to a comma", then the three loader call sites (pandas, polars, spark) routed through `infer_delimiter`. Restore `.tsv` to `_EXTENSION_FORMATS` at the same time.

### Acceptance criteria

- `--left a.tsv --right b.tsv --on id` works with no extra flags
- `--left a.csv --right b.tsv --on id` works with no extra flags
- An explicit `--csv-delimiter ';'` still applies to both sides
- `.csv` vs `.csv` behaviour is unchanged
- Remove the `.. note::` in `docs/source/cli.rst` that documents the current limitation, and the `test_tsv_extension_is_not_inferred` test

### Out of scope

[...]

PR #549, opened 2026-08-03
fix(cli): infer delimiters for TSV inputs

AtomicGlance:
Addresses #545.

## What changed

- Infer comma or tab delimiters separately for each input file.
- Recognize `.tsv` and `.tab` as CSV inputs across pandas, Polars, and Spark.
- Keep `--csv-delimiter` as an override for both inputs.
- Replace the old TSV limitation test with same-format and mixed CSV/TSV coverage, including a Spark regression test.
- Update the CLI documentation to describe the inferred extensions.

## Why

The CLI already inferred each file's format independently, but the delimiter was one shared value that defaulted to a comma. A TSV file could therefore reach the correct reader and still be parsed into one malformed column. Keeping the parser default unset lets the loaders distinguish an explicit override from normal extension-based inference.

## Validation

- `pytest tests/cli/test_compare.py tests/cli/test_parser.py tests/cli/test_spark.py -q` — 124 passed, 5 skipped
- `ruff check datacompy/cli tests/cli`
- `ruff format --check datacompy/cli tests/cli`
- `mypy --follow-imports=skip --ignore-missing-imports datacompy/cli/backends.py`
- `sphinx-build -b html docs/source docs/build/html` (builds with the existing `api/modules` toctree warning)

### DIFF
--- a/datacompy/cli/backends.py
+++ b/datacompy/cli/backends.py
@@ -48,19 +48,18 @@ from datacompy.cli.parser import OPTIONS
 
 #: File extension to canonical format name.
 #:
-#: ``.tsv`` is deliberately absent. Mapping it to ``csv`` would pick the right
-#: reader but not the right delimiter, which stays comma unless
-#: ``--csv-delimiter`` says otherwise, so a ``.tsv`` file would be recognised
-#: and then misparsed. Until the delimiter is inferred per file, a tab
-#: separated file is read with an explicit ``--input-format csv``.
 _EXTENSION_FORMATS = {
     ".csv": "csv",
+    ".tab": "csv",
+    ".tsv": "csv",
     ".parquet": "parquet",
     ".pq": "parquet",
     ".json": "json",
     ".jsonl": "json",
     ".ndjson": "json",
 }
+
+_DELIMITER_EXTENSIONS = {".tab": "\t", ".tsv": "\t"}
 
 _NDJSON_EXTENSIONS = frozenset({".jsonl", ".ndjson"})
 
@@ -98,6 +97,18 @@ def infer_format(ref: str, override: str | None) -> str:
             f"cannot infer the format of {ref!r} from its extension "
             f"{extension or '(none)'!r}. Pass --input-format csv|parquet|json."
         ) from None
+
+
+def infer_delimiter(ref: str, override: str | None) -> str:
+    """Return the field delimiter for *ref*.
+
+    An explicit ``--csv-delimiter`` takes precedence over extension based
+    inference. CSV files and paths with no recognised delimiter extension use
+    a comma.
+    """
+    if override is not None:
+        return override
+    return _DELIMITER_EXTENSIONS.get(Path(ref).suffix.lower(), ",")
 
 
 def _is_ndjson(ref: str) -> bool:
@@ -203,7 +214,9 @@ class PandasBackend(CLIBackend):
         fmt = infer_format(ref, namespace.input_format)
         try:
             if fmt == "csv":
-                return pd.read_csv(ref, sep=namespace.csv_delimiter)
+                return pd.read_csv(
+                    ref, sep=infer_delimiter(ref, namespace.csv_delimiter)
+                )
             if fmt == "parquet":
                 return pd.read_parquet(ref)
             return pd.read_json(ref, lines=_is_ndjson(ref))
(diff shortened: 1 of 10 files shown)


fdosani:
Thanks for the PR. I'd like to hold off merging this in right now as I want to focus on getting the original PR reviewed and merged. I'll leave it be for now, but maybe get you to do a rebase/alteration once we are ready to move forward with this.

fdosani (commented):
Taking another pass at the review. Sorry about the confusion

datacompy/cli/backends.py
```diff
@@ -92,14 +100,96 @@ def infer_format(ref: str, override: str | None) -> str:
...
+    frame_columns = getattr(frame, "columns", None)
+    if frame_columns is None:
+        return None
+    columns = list(frame_columns)
+    if len(columns) != 1:
+        return None
+
+    name = str(columns[0])
+    used = infer_delimiter(ref, namespace.csv_delimiter)
+    if not any(char != used and char in name for char in _PLAUSIBLE_DELIMITERS):
```
fdosani:
```suggestion
    # A collapsed multi-column header repeats the real delimiter once per lost
    # column, so a plausible delimiter appearing more than once is the mark of a
    # misparse. A single occurrence is just as easily a genuine column name such
    # as "a;b", so it is left alone rather than warned about on a correct run.
    if not any(char != used and name.count(char) > 1 for char in _PLAUSIBLE_DELIMITERS):
```

datacompy/cli/backends.py
```diff
@@ -92,14 +100,96 @@ def infer_format(ref: str, override: str | None) -> str:
...
+        The loaded DataFrame. Only ``columns`` is read, which every file
+        backend exposes without touching the data.
+
+    Returns
+    -------
+    str or None
+        A warning message, or ``None`` when the parse looks fine. Only a single
+        column CSV whose one column name contains a delimiter other than the
+        one used is reported, so a genuine single column file and a plain typo
+        in ``--on`` stay quiet.
```
fdosani:
```suggestion
        A warning message, or ``None`` when the parse looks fine. Only a single
        column CSV whose one column name *repeats* a delimiter other than the
        one used is reported, because that is what a collapsed multi-column
        header looks like. A genuine single column file, a name that merely
        contains one such delimiter (for example ``a;b``), and a plain typo in
        ``--on`` all stay quiet.
```

tests/cli/test_backends.py
```diff
@@ -0,0 +1,178 @@
...
+def test_a_quoted_header_is_not_reported_as_a_misparse(frame_of):
+    """A single column genuinely named ``a,b`` survives a comma delimited read.
+
+    The comma in that name is the delimiter the file was read with, so it
+    cannot be evidence that the wrong one was chosen. Without that check the
+    only correctly parsed file in the world with a comma in a column name gets
+    warned about on every run.
+    """
+    frame = frame_of("a,b")
+    assert suspect_delimiter("data.csv", _namespace(), frame) is None
```
fdosani:
```suggestion
@pytest.mark.parametrize(
    ("ref", "csv_delimiter", "column_name", "warns"),
    [
        # Read with a comma (.csv): a lone foreign delimiter is an ordinary
        # name...
        ("data.csv", None, "a;b", False),
        ("data.csv", None, "a|b", False),
        ("data.csv", None, "a\tb", False),
        # ...but a repeated one is a collapsed multi-column header.
        ("data.csv", None, "a;b;c", True),
        ("data.csv", None, "a|b|c", True),
        ("data.csv", None, "a\tb\tc", True),
        # The delimiter actually used never implicates itself, however often it
        # appears (a quoted "a,b,c" survives a comma read as one column).
        ("data.csv", None, "a,b", False),
        ("data.csv", None, "a,b,c", False),
        # Two different foreign delimiters, one each, stays ambiguous -> quiet.
        ("data.csv", None, "a;b|c", False),
        # Nothing plausible in the name.
        ("data.csv", None, "id", False),
        ("data.csv", None, "amount_usd", False),
        # Read with a tab (.tsv): the mirror image.
        ("data.tsv", None, "first,last", False),
        ("data.tsv", None, "id,name,amount", True),
        ("data.tsv", None, "a;b", False),
        ("data.tsv", None, "a;b;c", True),
        ("data.tsv", None, "a\tb\tc", False),  # tab is what it was read with
        # An explicit --csv-delimiter is what "used" means, not the extension.
        ("data.csv", ";", "id,name,amount", True),
        ("data.csv", ";", "a;b;c", False),  # ';' is what it was read with
        ("data.csv", ";", "a|b", False),
        ("data.csv", ";", "a|b|c", True),
    ],
)
def test_suspect_delimiter_decision_matrix(
    frame_of, ref, csv_delimiter, column_name, warns
):
    """A single column is a misparse only when a foreign delimiter repeats.

    A plausible delimiter that appears more than once is what a collapsed
    multi-column header leaves behind. A lone occurrence is just as easily a
    genuine column name, and the delimiter the file was read with never
    implicates itself. Parametrised over both in memory backends via
    ``frame_of``.
    """
    namespace = _namespace(csv_delimiter=csv_delimiter)
    message = suspect_delimiter(ref, namespace, frame_of(column_name))
    assert (message is not None) == warns
```

PR #548, opened 2026-08-02
fix: support Spark Connect sessions

patrickswedish:
## Summary

- dispatch Spark SQL functions and window expressions to the implementation matching the actual classic or Connect DataFrame/Column
- apply the same dispatch throughout `SparkSQLCompare` and its built-in array, boolean, numeric, and string comparators
- add regression coverage proving Connect expressions can be built without an active local `SparkContext`

## Validation

- `ruff check` and `ruff format --check` on every changed Python file
- Spark Connect regression tests: passed
- Spark comparator and helper suite: 34 passed
- focused end-to-end `SparkSQLCompare` paths (merge, duplicate handling, mismatch reporting, windows, masking, and cache behavior): 8 passed

Closes #535

PR #552, opened 2026-08-06
fix: support Spark Connect sessions in SparkSQLCompare

### LINKED ISSUE #535: SparkSQLCompare fails with Spark Connect sessions (pyspark.sql.connect.session.SparkSession)
### Summary
datacompy.spark.SparkSQLCompare raises an AssertionError when used with a Spark Connect session (pyspark.sql.connect.session.SparkSession). It only works with a classic JVM-backed pyspark.sql.SparkSession.

### Steps to Reproduce
```
# spark is a pyspark.sql.connect.session.SparkSession (e.g. from Spark Connect)  
import datacompy  
  
comp = datacompy.spark.SparkSQLCompare(  
    spark, df1, df2, join_columns="id"  
)  
```

### Error
```
AssertionError:  
  File ".../datacompy/spark.py", line 440, in _dataframe_merge  
    df1 = df1.withColumn("_merge_left", F.lit(True))  
  File ".../pyspark/sql/functions.py", line 95, in _invoke_function  
    assert SparkContext._active_spark_context is not None 
```

### Root Cause
datacompy/spark.py imports from pyspark.sql.functions (the classic JVM-backed module), which internally asserts that a local SparkContext is active. Spark Connect sessions have no local SparkContext — they communicate with the Spark driver over a remote gRPC connection — so this assertion always fails.

### Expected Behavior
SparkSQLCompare should work with both classic pyspark.sql.SparkSession and Spark Connect pyspark.sql.connect.session.SparkSession.

fdosani:
## Summary

`SparkSQLCompare` did not work against a Spark Connect session it was handed rather than one it created (issue #535). The root cause is that `pyspark.sql.functions` and `pyspark.sql.Window` only forward to the Spark Connect implementations while the process-global `SPARK_CONNECT_MODE_ENABLED` environment variable is set. A notebook runtime, a serverless runtime, or another framework that hands over an already-built Connect session does not necessarily set it, so datacompy built classic expressions against Connect DataFrames.

This branch selects the expression API from the object being operated on instead of from the global flag, and then backs that up with a Connect test session, a regression suite, and CI/tox coverage so the path stays green.

## Credit

The dispatch approach here is the one proposed by @patrickswedish in #548, and that PR is what identified the process-global flag as the real cause. This branch keeps that design and extends it:

- `is_spark_connect_object()` walks the MRO comparing module names rather than doing `isinstance` against the Connect classes, so a classic-only PySpark install never triggers the `pyspark.sql.connect` import that requires the optional `grpcio` dependency
- fixes to the non-comparator Connect breakages found while building an end-to-end Connect test run (report rendering, temp views, empty `withColumns`)
- a Spark Connect test session, a regression suite, and CI + tox coverage

Closes #535. Supersedes #548.

## Library changes

**Expression dispatch** (`datacompy/comparator/utility.py`)

- `is_spark_connect_object()`, `get_spark_functions()`, and `get_spark_window()` resolve the classic or Connect `functions` module and `Window` class from the DataFrame or Column at hand
- every call site in `datacompy/spark.py` and in the array, boolean, numeric, and string comparators resolves `F` / `Window` this way instead of importing them at module scope

**Other Connect breakages** (`datacompy/spark.py`, `datacompy/base.py`)

- `df_to_str()` checked `hasattr(df, "to_string")` before `hasattr(df, "toPandas")`. A Connect DataFrame synthesizes a Column for any unknown attribute, so the first check is always True for it and every Spark report took the pandas branch. The `toPandas` check now comes first.
- `_dataframe_merge()` registered fixed `df1` / `df2` temp views. Connect resolves views lazily, so a second comparison on the same session replaced the views an earlier plan still referred to; the names are now unique per merge, which also stops datacompy clobbering a user view called `df1` or `df2`.
- `_intersect_compare()` now guards `withColumns()` on a non-empty mapping. Connect asserts on an empty one where classic Spark treats it as a no-op; this is reached when the two frames share nothing but the join columns.
- the type check in `_validate_dataframe()` no longer imports `pyspark.sql.connect.dataframe`, for the same `grpcio` reason as above.

## Tests

- `tests/test_spark_connect.py`: end-to-end `SparkSQLCompare` coverage against a real Connect session. The session fixture clears `SPARK_CONNECT_MODE_ENABLED`, which is what makes these tests able to fail: with the flag left set, `pyspark.sql.functions` forwards to Connect on its own and the suite would pass against the unfixed library.
- `tests/comparator/test_utility_spark.py`: unit coverage for the three new helpers.
- `pytest-connect.ini` runs the existing `tests/test_spark.py` and `tests/comparator/` against a Connect session, so the fix is checked against the full existing suite rather than only the new one.

The Connect suites must each run in their own pytest process: starting a local Connect server sets `SPARK_LOCAL_REMOTE`, after which every later `SparkSession.builder.getOrCreate()` in that process returns the Connect session, so classic and Connect sessions cannot coexist in one run. The regression suite is marked `spark_connect` and deselected from the default run via `addopts`.

## CI and tooling

- `.github/workflows/test-package.yml`: the Spark matrix was a cross product where the Spark jobs are the entire cost of a run (~20 min each) and the no-Spark jobs finish in under a minute. It now covers the axes independently: Python 3.10-3.13 breadth from the four cheap basic-install jobs, pandas 2 vs 3 and Spark 3.5 vs 4 each present once, and ANSI mode plus the two Spark Connect steps on the 3.12 baseline only. The known gap (no Spark job on 3.11 or 3.13) is recorded in the file's header along with what closing it costs.
- `tox.ini` (new): mirrors that workflow job for job via `tox-conda`, so the CI matrix can be reproduced locally without a system JDK. Its header documents two traps that cost real time to find: the `openjdk>=17.0.8,<18` floor (below it, conda-forge selects a GraalVM build that silently replaces CPython with GraalPy) and `{envpython}` substituting to the string `None` in tox 3.28.
- `Makefile`: `make test`, `test-ansi`, `test-connect`, `test-connect-regression`, `test-cov`, `test-all`, plus `-no-snowflake` variants.
- `pytest.ini` / `pytest-ansi.ini` / `pytest-connect.ini`: all three now set `testpaths = tests`. Without it a bare `pytest` walks the whole repo, and a worktree checked out under the repo root contributes a second `tests/conftest.py`, which aborts the run with `ImportPathMismatchError`. The Spark driver and the Connect server are also bound to `127.0.0.1` so tests do not depend on the host's outward-facing network config.
- `pyproject.toml`: `ruff>=0.16` on the `qa` extra, since the config uses selectors that older ruff fails to parse; dropped the Python 3.14 classifier, which claimed support for a version nothing in the test matrix covers.

## Docs

- `docs/source/spark_usage.rst`: a Spark Connect section covering the handed-over-session case and the reason dispatch is per-DataFrame, plus notes on the `connect` extra, `cache_intermediates=False` on runtimes that forbid caching, and Arrow always being on the wire under Connect
- `docs/source/install.rst`: what `datacompy[spark]` installs and what a self-managed PySpark needs for Connect
- `README.md`: Spark listed as classic and Spark Connect
- `CLAUDE.md`: the testing, tox, and dispatch notes above

## Validation

CI covers this: the 3.12 Spark 4 job runs the default suite, ANSI mode, the existing Spark suite against a Connect session, and the Connect regression suite. Locally, `make test-all` runs the same set.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -1,5 +1,23 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+#
+# Matrix design. The Spark jobs are the entire cost of this workflow (~20 min
+# each); lint and the four basic-install jobs finish in under a minute. So the
+# test axes are covered independently rather than as a cross product:
+#
+#   * python 3.10-3.13 breadth comes from the four cheap basic-install jobs,
+#     which exercise all the pandas/polars/base/report/comparator code.
+#   * pandas 2 vs 3 is a real API split, so both appear -- pandas 2.3.3 on the
+#     python 3.10 Spark job (3.10 cannot take pandas 3), pandas 3.0.3 on 3.12.
+#   * spark 3.5 vs 4 is a real API split, so both appear.
+#   * ANSI mode and Spark Connect are spark-side semantics, orthogonal to the
+#     python and pandas versions, so they run once on the 3.12 baseline rather
+#     than on every Spark job.
+#
+# Known gap: no Spark job runs on python 3.11 or 3.13, so a Spark break
+# specific to those runtimes would not be caught. Widening is one entry in the
+# test-with-spark-4-install include list (and one line in tox.ini's envlist,
+# which mirrors this file job for job).
 
 name: Test package
 
@@ -29,29 +47,18 @@ jobs:
       - name: Formatting by ruff
         run: ruff format --check
 
+  # Legacy Spark path. One job is enough to prove 3.5 still works; the wider
+  # combinations remain available locally via `tox -e py310-spark35-pandas2`
+  # and friends.
   test-with-spark-3-install:
+    name: spark 3.5 (py3.11, pandas 2.3.3)
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: ["3.10", "3.11"]
-        spark-version: [3.5.8]
-        pandas-version: ["2.3.3", "3.0.3"]
-        exclude:
-          - python-version: "3.10"
-            pandas-version: "3.0.3"
-
-    env:
-      PYTHON_VERSION: ${{ matrix.python-version }}
-      SPARK_VERSION: ${{ matrix.spark-version }}
-      PANDAS_VERSION: ${{ matrix.pandas-version }}
-
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python 3.11
         uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6.3.0
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.11"
       - name: Setup Java JDK
         uses: actions/setup-java@ad2b38190b15e4d6bdf0c97fb4fca8412226d287  # v5.3.0
         with:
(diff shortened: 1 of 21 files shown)


patrickswedish:
Tested the dispatch approach end-to-end locally against a Connect session — the `is_spark_connect_object()` MRO walk is clean, the `grpcio` import guard works, and the temp-view uniqueness fix is solid. Happy to help with any follow-up review items if the maintainers have questions. 🚀

rhaffar (approved):
Looks good Faisal - for my understanding, these new utilities mostly exist to support spark sql and spark connect for versions older than 4.0 correct? Because from what I'm reading, the API is entirely identical between Spark Classic and Connect from 4.0 onwards and the implementation can switch between the two by setting a flag. If that is the case, maybe we want to consider deprecating versions older than 4.0 at some point in the future.

fdosani:
> Looks good Faisal - for my understanding, these new utilities mostly exist to support spark sql and spark connect for versions older than 4.0 correct? Because from what I'm reading, the API is entirely identical between Spark Classic and Connect from 4.0 onwards and the implementation can switch between the two by setting a flag. If that is the case, maybe we want to consider deprecating versions older than 4.0 at some point in the future.

The name space is different between connect and non-connect. They should function similarly!
I think at some point yes we will want to deprecate 3.x and just move to 4.x.

PR #559, opened 2026-08-30
fix: name the delimiter as a likely cause when join columns are missing from a CSV input

### LINKED ISSUE #553: CLI: name the delimiter as a likely cause when join columns are missing from a CSV input
Follow-up from #545 and #549.

### Context

When a CSV family input is read with the wrong delimiter, every row collapses into one column whose name is the entire header line. The comparison then fails on the join columns, and the only thing the user sees is the library's message:

```
$ datacompy compare --left left.tsv --right right.tsv --on id --backend pandas
datacompy: df1 must have all columns from join_columns: {'id'}
```

That message is correct and points nowhere near the cause. `id` really is absent, because the frame has a single column named `id,name,amount`. The user is left looking at their join columns rather than at the delimiter.

#545 makes this rarer by inferring the delimiter from the extension, but it also removes the guard that used to catch the case. Before that change a `.tsv` input without `--input-format` failed with an actionable "cannot infer the format" error. Afterwards, a comma delimited file that happens to be named `.tsv`, or a tab delimited file with no distinguishing extension, lands on the message above instead. Under `--on-index` there is no error at all: the comparison runs on one mangled string column per side and reports on that.

The message comes from the comparison classes (`datacompy/pandas.py:262`, `datacompy/polars.py:254`, `datacompy/spark.py:322`) as a plain `ValueError`, and `datacompy/cli/compare.py:62` translates it into a `BadArgsError` verbatim.

### What to do

Keep the library messages as they are and add the hint in the CLI, which is the only layer that knows the input was a file, what extension it had, and which delimiter was chosen for it.

When the missing join columns error fires, and the offending side was read as `csv`, and the resulting frame has exactly one column whose name contains a plausible delimiter (`,`, `\t`, `;`, `|`), append a second line:

```
datacompy: df1 must have all columns from join_columns: {'id'}
Hint: 'left.tsv' parsed into a single column named 'id,name,amount'. The
delimiter may be wrong. Pass --csv-delimiter to override the extension based
default.
```

The `ValueError` is raised inside `backend.build`, well after the loaders have run, so the loaders need to record what they did. A small per side record (reference, resolved format, resolved delimiter, column count and names) carried through `run()` is enough, and all four backends expose `.columns` on the frame they return.

The same check is worth applying on the `--on-index` path, where nothing raises today. A single column frame on either side whose name contains a plausible delimiter should produce a warning on stderr before the report, since a silent comparison of two mangled strings is the worst version of this failure.

### Acceptance criteria

- A comma delimited file named `.tsv` compared with `--on id` reports the missing join column *and* the delimiter hint, naming the offending input
- The hint names the file whose parse looks wrong, not both sides, when only one side is affected
- A genuine missing join column on a correctly parsed file, for example `--on nope` against a normal `.csv`, is unchanged and gains no hint
- A legitimate single column CSV whose header contains no delimiter character gains no hint
- The `--on-index` path warns on stderr when either side parsed into a single suspicious column, and still exits on the comparison result rather than failing
- Library exception messages and types are unchanged, so this stays a CLI concern

### Out of scope

Sniffing the delimiter from file content, either to guess it or to validate the extension based choice. The hint only reads what was already parsed. Detecting the delimiter properly is a larger change and would need its own discussion about precedence against `--csv-delimiter` and the extension.

BhargavKumarNath:
Closes #553.

## What

When a CSV-family input is read with the wrong delimiter, every row collapses into a single column whose name is the whole header line. The comparison then fails on the join columns with a message that is correct but points nowhere near the cause:

$ datacompy compare --left left.csv --right right.csv --on id --backend pandas
datacompy: df1 must have all columns from join_columns: {'id'}

This adds a hint at the CLI layer, which is the only layer that knows the input was a file, what extension it had, and which delimiter was chosen. Library messages and exception types are untouched.

## Behaviour

**`--on` path** — when the side named in the error was read as CSV and collapsed to a single column whose name still contains a plausible delimiter (`,`, `\t`, `;`, `|`):

datacompy: df1 must have all columns from join_columns: {'id'}
Hint: 'left.csv' was read as CSV with delimiter ',' and collapsed to a single
column named 'id\tname\tamount'. The delimiter is probably wrong; set it with
--csv-delimiter.

The hint names only the offending side. A genuine missing column on a correctly parsed file, and a legitimate single-column CSV whose header has no delimiter character, get no hint.

**`--on-index` path** — nothing raises today; the comparison runs silently on two mangled string columns. Now a single suspicious column on either side produces a warning on stderr before the report, and the exit code still reflects the comparison result.

## How

- `ParsedInput` (new, `datacompy/cli/backends.py`): a per-side record of what `load()` did (reference, resolved format, resolved delimiter, column names) with a `looks_misparsed` predicate. `CLIBackend.describe()` builds it after a successful load and returns `None` for the snowflake backend, whose inputs are table references rather than parsed files.
- `datacompy/cli/compare.py`: `run_compare` carries the two records through; `add_delimiter_hint()` augments the `ValueError` message on the `--on` path, `warn_on_misparsed_index_join()` covers `--on-index`.
- `print_warning()` (new, `datacompy/cli/output.py`).
- Doc paragraph in `docs/source/cli.rst`.

## Tests

13 cases added to `tests/cli/test_compare.py`: the hint across `\t` / `;` / `|` on both in-memory backends, the hint naming only the offending side, no hint for a real missing column, no hint for a delimiter-free single-column CSV, and the `--on-index` stderr warning.

- `pytest tests/cli` and `pytest tests/test_pandas.py tests/test_polars.py tests/test_base.py` pass
- `ruff check`, `ruff format --check`, `pre-commit run --files ...` clean
- `mypy datacompy/cli` introduces no new errors against the baseline

### DIFF
--- a/datacompy/cli/backends.py
+++ b/datacompy/cli/backends.py
@@ -36,6 +36,7 @@ import re
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from contextlib import ExitStack
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, ClassVar
 
@@ -103,6 +104,41 @@ def infer_format(ref: str, override: str | None) -> str:
 def _is_ndjson(ref: str) -> bool:
     """Return ``True`` when *ref* looks like newline delimited JSON."""
     return Path(ref).suffix.lower() in _NDJSON_EXTENSIONS
+
+
+#: Delimiter characters worth naming when a CSV parse collapses to one column.
+#: A header line read with the wrong separator keeps whichever of these it was
+#: meant to be split on.
+_SUSPICIOUS_DELIMITERS = (",", "\t", ";", "|")
+
+
+@dataclass(frozen=True)
+class ParsedInput:
+    """What a file backend's :meth:`CLIBackend.load` did with one input.
+
+    Recorded per side and carried through
+    :func:`datacompy.cli.compare.run_compare` so that a later join column
+    failure can point at a delimiter mistake rather than leave the user
+    studying their join keys. The snowflake backend produces ``None`` instead,
+    since its inputs are table references, not parsed files.
+    """
+
+    ref: str
+    fmt: str
+    delimiter: str | None
+    columns: tuple[str, ...]
+
+    @property
+    def looks_misparsed(self) -> bool:
+        """Whether this looks like a CSV read with the wrong delimiter.
+
+        The tell is a CSV input that collapsed to a single column whose name
+        still carries a plausible delimiter, which is exactly the shape of a
+        header line split on the wrong character.
+        """
+        if self.fmt != "csv" or len(self.columns) != 1:
+            return False
+        return any(char in self.columns[0] for char in _SUSPICIOUS_DELIMITERS)
 
 
 def compare_kwargs(namespace: argparse.Namespace, backend: str) -> dict[str, Any]:
@@ -181,6 +217,27 @@ class CLIBackend(ABC):
     @abstractmethod
     def load(self, session: Any, ref: str, namespace: argparse.Namespace) -> Any:
         """Turn *ref* into whatever the backend's comparison accepts."""
+
+    def describe(
+        self, ref: str, namespace: argparse.Namespace, payload: Any
+    ) -> ParsedInput | None:
+        """Record how :meth:`load` read *ref*, for diagnostics after the fact.
+
+        Returns ``None`` when *payload* is not a parsed frame, which is the case
+        for the snowflake backend where *payload* is a table name. Called only
+        after :meth:`load` has succeeded, so :func:`infer_format` cannot raise
+        here.
+        """
+        columns = getattr(payload, "columns", None)
+        if columns is None:
+            return None
+        fmt = infer_format(ref, namespace.input_format)
+        return ParsedInput(
+            ref=ref,
+            fmt=fmt,
+            delimiter=namespace.csv_delimiter if fmt == "csv" else None,
+            columns=tuple(str(col) for col in columns),
+        )
 
     def build(
         self, namespace: argparse.Namespace, session: Any, left: Any, right: Any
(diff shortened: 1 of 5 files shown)


BhargavKumarNath:
This is ready for review.

Just one note on the ordering: #553 was split out from the #545 review, so there’s a small dependency between the two. I’ve used “set it with --csv-delimiter” in the hint for now, since the per-file delimiter inference from #545 isn’t on `main` yet. The wording can be changed to match the issue once #545 is merged.

If #545 lands first, I’ll make that small wording change.

fdosani:
Hi there there is already an existing PR for this issue #549 . Closing in favour for it. If you have any suggestions you want to include please make a comment on the existing PR / Issue.

BhargavKumarNath:
Hi @AtomicGlance, #549 was pointed here after my #559 was closed as a duplicate, so I wanted to pass along one small suggestion I had.

I think this covers the delimiter inference and pre-build warning really nicely. One small thing I noticed against the #553 acceptance criteria is that `suspect_delimiter` currently prints the warning separately before `backend.build`. If the join then fails, the actual error is still:

`datacompy: df1 must have all columns from join_columns: {'id'}`

So in a longer run, the hint could easily get lost above the error. The #553 example instead has the hint attached directly to the error.

Would it make sense to also add the hint when converting this particular `ValueError` to `BadArgsError` in `run_compare`? Something along these lines, reusing `suspect_delimiter`:

```python
except ValueError as exc:
    message = str(exc)
    match = _MISSING_JOIN_COLUMNS.match(message)
    if match is not None:
        side = _SIDE_FOR[match.group(1)]
        ref = namespace.left if side == "left" else namespace.right
        frame = left if side == "left" else right
        suspect = suspect_delimiter(ref, namespace, frame)
        if suspect is not None:
            message = f"{message}\nHint: {suspect}"
    raise BadArgsError(message) from exc
```

That would keep the existing standalone warning for the `--on-index` path, while also putting the hint directly on the error for column joins.

If you think that makes sense, I'm happy to send a small follow-up PR for it, either against your branch or after #549 merges, whichever is easier.

PR #563, opened 2026-09-15
fix(comparator): handle all-null object columns in string comparison

### LINKED ISSUE #562: PandasCompare reports a mismatch on an all-null object-dtype column even when both sides are identical
Comparing two identical DataFrames reports a mismatch on any object-dtype column that is null on every row in both frames.

Repro (datacompy 1.0.4):

```python
import pandas as pd
import datacompy

df1 = pd.DataFrame({'id': [1, 2, 3], 'note': [None, None, None]})
df2 = pd.DataFrame({'id': [1, 2, 3], 'note': [None, None, None]})

c = datacompy.PandasCompare(df1, df2, join_columns=['id'])
c.matches()
```

Expected `True`, the frames are identical. Actual `False`. The per-column stats for `note` show `null_diff: 0` (both sides really are null everywhere) but `unequal_cnt: 3` and `all_match: False`.

Same thing through the CLI: two identical parquet files with an all-null object column, compared with `python -m datacompy compare --left left.parquet --right right.parquet --on id --backend pandas`, report the column as differing on every row.

For contrast, an all-NaN numeric column and an all-null nullable boolean column both compare equal correctly, so this looks specific to the object/free-text dtype path.

Tested on a clean clone at HEAD 7cb4005, `pip install -e .`, no other changes.

fdosani:
Fixes #562 

## What this is for
Comparing two identical tables was wrongly reporting a difference whenever a text column was empty (all blank/null) on both sides. This fix makes that case compare correctly as a match.

## What it does
- Teaches the string comparison logic to recognize an all-empty column instead of giving up and calling it a mismatch
- Adds tests covering two all-empty columns, and one empty column compared against a column that has some real text in it


rhaffar (approved):
LGTM

PR #564, opened 2026-09-16
Fix #551 : add DataComPy agent skill (SKILL.md, backends, migration guide)

### LINKED ISSUE #551: Add a DataComPy agent skill (SKILL.md + references), modeled on polars-inc/skills
## Summary

[polars-inc/skills](https://github.com/polars-inc/skills/tree/main/polars) ships a first-party **agent skill** for Polars: a `SKILL.md` plus on-demand `references/` files that teach coding agents (Claude Code, Codex, Copilot, Cursor) how to use the library correctly, following the [SKILL.md open standard](https://code.claude.com/docs/en/skills).

DataComPy is a good candidate for the same treatment. Agents writing DataComPy code today tend to reach for the v0 `datacompy.Compare` API, pick the wrong backend class, misuse `abs_tol`/`rel_tol`, or hand-roll mismatch analysis that `all_mismatch()` / `report()` already do. A skill fixes that at the source instead of one prompt at a time.

## Proposed layout

```
skills/datacompy/
├── .claude-plugin/
│   └── plugin.json          # Claude Code plugin manifest
├── SKILL.md                 # core methodology + routing to references
└── references/
    ├── backends.md
    ├── joins-and-keys.md
    ├── tolerances.md
    ├── reports.md
    ├── comparators.md
    ├── migration-v0-to-v1.md
    └── troubleshooting.md
```

No build step; the directory is the deliverable. It lives in this repo so it versions alongside the API it documents.

## SKILL.md core content

Frontmatter `name` / `description` written so agents auto-load it on "compare two dataframes", "diff these tables", "reconcile", "PROC COMPARE", "datacompy".

The prescribed workflow:

1. **Pick the backend class by dataframe type, not by preference**: `PandasCompare`, `PolarsCompare`, `SparkSQLCompare`, `SnowflakeCompare`. Never convert a Spark or Snowpark dataframe to Pandas just to compare it.
2. **Establish the join key before anything else.** `join_columns` (or `on_index=True` for Pandas/Polars only). Verify uniqueness of the key first, since duplicate keys change the semantics of the whole comparison.
3. **Construct once, then read results off the object.** Don't recompute; `_compare` runs at init.
4. **Answer with the built-ins**: `matches()`, `subset()`, `all_columns_match()`, `all_rows_overlap()`, `intersect_rows_match()`, `count_matching_rows()`, `columns_with_mismatches()`, `only_join_columns()`, `sample_mismatch()`, `all_mismatch()`, `report()`, `build_report_data()`.
5. **Set tolerances deliberately.** Default is exact (`abs_tol=0`, `rel_tol=0`); floats almost always need one.

Core rules to encode:

- `matches()` is the boolean; `report()` is for humans. Don't parse the report string to make a decision, use the accessors or `build_report_data()`.
- `cast_column_names_lower=True` is the default on Pandas/Polars; Snowflake normalizes to **upper**. Tolerance dict keys must match that casing. This is a real footgun and gets its own callout.
- `abs_tol`/`rel_tol` accept a float *or* a `dict[str, float]` for per-column tolerances (`validate_tolerance_parameter` in `base.py`).
- `sensitive_columns` masks values in the report; `reveal_sensitive_columns()` unmasks. Mention it before an agent writes its own redaction.
- Spark and Snowflake are optional imports. Guard on `ImportError`, and never assume `datacompy.SparkSQLCompare` exists.
- `on_index` and `join_columns` are mutually exclusive, and `on_index` is not available on all backends.

Gotchas worth calling out explicitly (the equivalent of the Polars skill's "string literals are column names" section):

- Column-name case folding silently renaming columns, and what that does to `join_columns`.
- `ignore_spaces` / `ignore_case` apply to string comparison, not to join-key matching.
- Unequal row counts vs. unequal *keys*: `all_rows_overlap()` vs. `matches()`.
- Comparing on a non-unique key produces a many-to-many merge, not an error.

## references/ contents, one file per axis

| File | Contents |
|---|---|
| `backends.md` | Class-per-backend table, constructor signature differences (Snowflake takes a `session` first and accepts table-name strings), install extras, which options each backend supports. |
| `joins-and-keys.md` | Choosing `join_columns`, `on_index`, composite keys, verifying uniqueness first, what happens when keys are duplicated or null. |
| `tolerances.md` | `abs_tol` vs `rel_tol` semantics, float vs dict form, per-column tolerances, casing rules for dict keys per backend. |
| `reports.md` | `report()`, `template_path=` for custom Jinja2 templates, `save_html_report()`, `build_report_data()` and the `ReportData`/`ColumnSummary`/`MismatchStats` types for programmatic consumption. |
| `comparators.md` | The `datacompy.comparator` subpackage: numeric / string / boolean / array-like per backend, and how to write and pass a `custom_comparators` list. |
| `migration-v0-to-v1.md` | `datacompy.Compare` to `PandasCompare`, `SparkCompare` to `SparkSQLCompare`, renamed/removed args. This is the highest-value file, since agents trained on pre-1.0 material default to the old API. |
| `troubleshooting.md` | Common errors mapped to fixes: "Only provide on_index or join_columns", missing-optional-dependency import errors, tolerance key mismatches, memory blowups on `all_mismatch()` for wide frames. |

## Open questions

- **Placement**: `skills/datacompy/` at repo root (mirrors the Polars layout, discoverable) vs. shipping inside the wheel. Leaning toward repo root, not packaged.
- **Scope**: one skill covering all four backends, or a core skill plus per-backend skills? Starting with one; the `references/` split already gives per-backend routing without the fragmentation.
- **Docs overlap**: `docs/source/*_usage.rst` covers much of this prose already. The skill should stay short and rule-shaped rather than duplicating the docs. Where a reference file and the docs would say the same thing, the reference file wins on brevity and links out.
- **Testing**: worth a small eval set (N prompts, expected API usage) to catch regressions when the API changes. Could start as a checklist rather than CI.
- **Branch**: should land on `develop` alongside v1, since the whole point is teaching the v1 API.

## Proposed sequencing

1. `SKILL.md` + `plugin.json` + `migration-v0-to-v1.md` + `backends.md`, covering the two failure modes that matter most (wrong API version, wrong backend class).
2. `tolerances.md`, `joins-and-keys.md`, `reports.md`.
3. `comparators.md`, `troubleshooting.md`.
4. README section on installing the skill; optional submission to skill directories.

prakhar2000:
Fix #551
## Summary

- Step 1 of #551: `skills/datacompy/SKILL.md`, `.claude-plugin/plugin.json`, `references/backends.md`, and `references/migration-v0-to-v1.md`, modeled on [polars-inc/skills](https://github.com/polars-inc/skills/tree/main/polars).
- Branched from `main` because CLAUDE.md treats `main` as the current release branch (the issue's `develop` note predates that).
- Written with AI assistance. Every behavior claim was checked against the code on `main`, and v0 names were checked against `support/0.19.x`, `v0.13.3`, and `v0.10.5`. Pandas and Polars examples were executed.

## Corrections to the #551 draft

The skill documents what the code does, not the issue text where they differ:

- Sensitive columns are `hide_sensitive_columns([...])` after construct. There is no `sensitive_columns=` constructor argument.
- Duplicate join keys are paired by row order within each key group, not merged many-to-many.
- `ignore_spaces` does strip join keys before matching. `ignore_case` does not; it only applies to compared string values after the join.
- Tolerance dict keys are normalized (lowercased when `cast_column_names_lower=True`, uppercased on Snowflake). A misspelled key is ignored silently; `AMT` vs `amt` is not the footgun once that runs. `on_index` is Pandas only, not Polars.

## Question, not a bug

`hide_sensitive_columns()` masks sample values, but the report's Max Diff column still prints the numeric difference (for example `29000` for salaries `50000` vs `79000`). Is that intentional, or should Max Diff be masked too when the column is hidden?

## Test plan

- [ ] `grep` for em dashes under `skills/` returns nothing
- [ ] Pandas and Polars code blocks in `SKILL.md` and `references/backends.md` run
- [ ] Confirm the skill never tells an agent to use `datacompy.Compare` or `sensitive_columns=`


skills/datacompy/SKILL.md
```diff
@@ -0,0 +1,163 @@
...
+  many-to-many. Two rows with `id=1` in each frame become two intersect
+  rows, first-with-first and second-with-second.
+- **`ignore_spaces` strips join keys before matching.
+  `ignore_case` does not.** `ignore_case` only affects compared string
+  values after the join.
+- **Null join keys match each other.** A null in df1 joins a null in
+  df2.
+- **`hide_sensitive_columns()` masks values in samples and unique-row
+  snippets, but `Max Diff` in the report still shows the numeric
+  difference.**
```
fdosani:
```suggestion
- **`hide_sensitive_columns()` does not fully protect a masked column.**
  Row-level values in samples and unique-row snippets are masked, but
  `Max Diff` in the report still shows the raw numeric difference. If
  one side of the comparison is already known, Max Diff can reveal the
  other. Treat masking as partial, not a full guarantee, and warn users
  who need to compare truly sensitive columns.
```

fdosani (changes requested):
Thanks for the PR @prakhar2000 

Confirmed this is a bug, not a documentation nuance. `hide_sensitive_columns()` only overwrites values in `df1_unq_rows`, `df2_unq_rows`, and `intersect_rows`. It never touches `column_stats` / `MismatchStat`, so `max_diff` (computed straight from raw values in `_intersect_compare()`) flows into the report unmasked. If one side of the comparison is already known publicly, Max Diff can expose the other side, which can defeat the purpose of masking.

I've filed this as a separate issue so it can be tracked and fixed independently of
this docs PR:  #565 

One small suggestion to the SKILL.md include

skills/datacompy/SKILL.md
prakhar2000: Thanks for the suggestion.
will apply the fix

fdosani:
@prakhar2000 the linting seems to be failing. Can you clean that up please when you have some time.

prakhar2000:
Fixed. ruff format --check wanted a blank line after the import in the first SKILL.md fence. Pushed.

fdosani (approved):

prakhar2000:
Thanks.
Just tag me if there is anything left on my side

